PR #2, opened 2020-05-14
FreeBSD updates

n-connect:
Several small changes in build/installation procedure. Original goal was to let run it on [Free]BSD, but in general is makes the scripts' run more flexible across Unix like and Linux distros.

### DIFF
--- a/scripts/get_continent_map.py
+++ b/scripts/get_continent_map.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3.7
+#!/usr/bin/env python3
 #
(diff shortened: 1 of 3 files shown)


scripts/get_continent_map.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3.7
```
go2sh, line 1: Thats not working for every one. Your distro should provide a symlink with the name `python` to the installed python version.

PR #7, opened 2020-05-20
Add Docker resources

gary-kim:
~~Still a work in progress.~~ Now complete.
This is the configuration I am using on my own server with some extra Janus command line flags and a separate Coturn server. I've made several calls with it and it works.

PR #14, opened 2020-05-26
add docker build to README, fixes #12

### LINKED ISSUE #12: Requirements for building
I just tried to clone and build the signaling server in Ubuntu 18.04 and found out there are at least to requirements which need to be installed first:

- golang
- curl

Maybe this should be added to the readme.

morph027:


fancycode:
Is this still required with #7 merged?

Chartman123:
I don't think that this is still needed... Especially if you consider that there has been no response from @morph027 for a long time now.

morph027:
Agree, #7 fixes this as well ;)

PR #15, opened 2020-05-28
Move processing of deferred method to dedicated DeferredExecutor.

fancycode:
This fixes an issue where re-joining a room after error stopped the internal processing of deferred functions.

@danxuliu please take a look to see if this fixes the issue you were having.

### DIFF
--- a/src/signaling/mcu_janus.go
+++ b/src/signaling/mcu_janus.go
@@ -371,7 +371,7 @@ type mcuJanusClient struct {
 	handle    *JanusHandle
 	handleId  uint64
 	closeChan chan bool
-	deferred  chan func()
+	deferred  *DeferredExecutor
 
 	handleEvent     func(event *janus.EventMsg)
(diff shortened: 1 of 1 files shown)


fancycode:
A whole `p.Close(ctx)` is not necessary as the subscriber is already partially closed at https://github.com/strukturag/nextcloud-spreed-signaling/blob/859ba9ec2cc1a1c7d6722b2e63f6782624f3f3b7/src/signaling/mcu_janus.go#L1011

I updated the code to properly cleanup the subscriber in case of errors so a new one will be created if the client retries the request. Can you confirm this is working for you with the latest commit?

Do you have the actual error from the `error during re-joining` message so I could maybe check the root cause for the error here? Thanks.

fancycode:
ping @danxuliu

fancycode:
Merging / closing for now, feel free to open a new issue if you experience a problem with this again in the future.

PR #26, opened 2020-07-07
Add "reason" field to disinvite message.

### LINKED ISSUE #25: Separate signaling messages for "session was removed" and "conversation was removed"
With https://github.com/nextcloud/spreed/pull/3591 we are improving the current world when people try to join twice in the same conversation. When your current session is removed a `disinvite` is sent and we show a message telling you:
> You joined the conversation in another window or device. This is currently not supported by Nextcloud Talk so this session was closed.

However we can not distinguish the disinvite from:
1. A moderator removed you from the conversation
2. You joined again and nextcloud send the HPB a "disinvite this session"
3. The conversation was removed completely.

Would it be possible to send different (or additional prior to the current one) messages so Talk can redirect you properly to a "The conversation does not exist" message?

fancycode:
This is required so clients can differentiate between no longer being invited or the room being deleted. Fixes #25.

cc @nickvergessen

### DIFF
--- a/src/signaling/api_signaling.go
+++ b/src/signaling/api_signaling.go
@@ -399,6 +399,17 @@ type RoomEventServerMessage struct {
 	Users   []map[string]interface{} `json:"users,omitempty"`
 }
 
+const (
+	DisinviteReasonDisinvited = "disinvited"
+	DisinviteReasonDeleted    = "deleted"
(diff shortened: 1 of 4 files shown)


fancycode:
Ping @nickvergessen

PR #28, opened 2020-07-21
Add support for multiple Nextcloud backends

fancycode:
Work in progress, needs some tests to make sure the isolation works.


fancycode:
Some isolation tests added.

PR #52, opened 2020-09-28
Feature: Reload Backends on SIGHUP

membero:
- Reload Backend Configuration without destroying unchanged backends
	- Only used when configured with backends parameter

### DIFF
--- a/src/signaling/backend_configuration.go
+++ b/src/signaling/backend_configuration.go
@@ -24,6 +24,7 @@ package signaling
 import (
 	"log"
 	"net/url"
+	"reflect"
 	"strings"
 
 	"github.com/dlintw/goconf"
@@ -70,40 +71,11 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration, 
 			compat: true,
 		}
 	} else if backendIds, _ := config.GetString("backend", "backends"); backendIds != "" {
-		seenIds := make(map[string]bool)
-		for _, id := range strings.Split(backendIds, ",") {
-			id = strings.TrimSpace(id)
-			if id == "" {
-				continue
+		for host, configuredBackends := range getConfiguredHosts(config) {
+			backends[host] = append(backends[host], configuredBackends...)
+			for _, be := range configuredBackends {
+				log.Printf("Backend %s added for %s", be.id, be.url)
 			}
-
-			if seenIds[id] {
-				continue
-			}
-			seenIds[id] = true
-
-			u, _ := config.GetString(id, "url")
-			secret, _ := config.GetString(id, "secret")
-			if u == "" || secret == "" {
-				log.Printf("Backend %s is missing or incomplete, skipping", id)
-				continue
-			}
-
-			if u[len(u)-1] != '/' {
-				u += "/"
-			}
-			parsed, err := url.Parse(u)
-			if err != nil {
-				log.Printf("Backend %s has an invalid url %s configured (%s), skipping", id, u, err)
-				continue
-			}
-
-			backends[parsed.Host] = append(backends[parsed.Host], &Backend{
-				id:     id,
-				url:    u,
-				secret: []byte(secret),
-			})
-			log.Printf("Backend %s added for %s", id, u)
 		}
 	} else if allowedUrls, _ := config.GetString("backend", "allowed"); allowedUrls != "" {
 		// Old-style configuration, only hosts are configured and are using a common secret.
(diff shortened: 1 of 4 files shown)


src/signaling/hub.go
```diff
@@ -420,6 +420,9 @@ func (h *Hub) Reload(config *goconf.ConfigFile) {
 	if h.mcu != nil {
 		h.mcu.Reload(config)
 	}
+	if h.backend != nil {
```
fancycode, line 423: The `h.backend` always exists, no need to check against `nil`.

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+			b.backends[host] = append(b.backends[host][:existingIndex], b.backends[host][existingIndex+1:]...)
+		}
+		existingIndex++
+	}
+
+	b.backends[host] = append(b.backends[host], backends...)
+}
+
+func (b *BackendConfiguration) Reload(config *goconf.ConfigFile) {
+	type configuredBackend struct {
```
fancycode, line 178: Unused, please remove.

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+	}
+	if backendIds, _ := config.GetString("backend", "backends"); backendIds != "" {
```
fancycode, line 180: This shares a lot of code with https://github.com/strukturag/nextcloud-spreed-signaling/blob/8a297b1a7067367b29aa070138f48db1620250a6/src/signaling/backend_configuration.go#L72:L107

Could you please move the common option handling / parsing code to a shared function?

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+		seenIds := make(map[string]struct{})
```
fancycode, line 181: Please simplify by using a `map[string]bool`.

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+		seenHosts := make(map[string]map[string]struct{})
```
fancycode, line 182: Please simplify by using a `map[string]map[string]bool`.

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+		for _, id := range strings.Split(backendIds, ",") {
+			id = strings.TrimSpace(id)
+			if id == "" {
+				continue
+			}
+
+			if _, ok := seenIds[id]; !ok {
+				seenIds[id] = struct{}{}
```
fancycode, line 190: Would become `seenIds[id] = true`.

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+			if u[len(u)-1] != '/' {
+				u += "/"
+			}
+			parsed, err := url.Parse(u)
+			if err != nil {
+				log.Printf("Backend %s has an invalid url %s configured (%s), skipping", id, u, err)
+				continue
+			}
+
+			if _, ok := seenHosts[parsed.Host]; !ok {
```
fancycode, line 208: This will only save the first id for a host. Also you could directly parse the `secret` and construct the `Backend` object here instead of parsing again below.

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+
+		// remove backends that are no longer configured
+		for hostname := range b.backends {
+			if _, ok := seenHosts[hostname]; !ok {
+				b.RemoveBackend(hostname)
+			}
+		}
+
+		// rewrite backends adding newly configured ones and rewriting existing ones
+		for hostname, host := range seenHosts {
```
fancycode, line 222: Naming: should be `for hostname, ids := range seenHosts` (value is a map of ids on that host).

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+			var backends []*Backend
+			for id := range host {
+				u, _ := config.GetString(id, "url")
```
fancycode, line 225: Duplicate config parsing (see comment above).

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+	delete(b.backends, host)
+}
+
+func (b *BackendConfiguration) UpsertHost(host string, backends []*Backend) {
+	existingIndex := 0
+	for _, existingBackend := range b.backends[host] {
+		found := false
+		index := 0
+		for _, newBackend := range backends {
+			if existingBackend == newBackend {
```
fancycode, line 161: This will compare object pointers which I think will always be different. Better would be to compare the URL of the backends.

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+				found = true
+				backends = append(backends[:index], backends[index+1:]...)
+				continue
```
fancycode, line 164: Shouldn't this be a `break`? You found the backend, so no need to further iterate the list.

Also: I don't see how a secret of an existing backend would be changed.

src/signaling/backend_configuration.go
```diff
@@ -148,6 +148,101 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
+		if !found {
+			b.backends[host] = append(b.backends[host][:existingIndex], b.backends[host][existingIndex+1:]...)
+		}
+		existingIndex++
+	}
+
+	b.backends[host] = append(b.backends[host], backends...)
+}
+
+func (b *BackendConfiguration) Reload(config *goconf.ConfigFile) {
```
fancycode, line 177: This should do nothing if a compatability configuration was loaded initially.

fancycode (changes requested):
Left some comments.

Please also create unittests in `backend_configuration_test.go` to make sure all possible scenarios are working as expected (add new backend on new host, add new backend to existing host, remove one of multiple backends from host, remove all backends from host, change url of backend, change secret of backend, + possible others).

membero:
Hi,
at the moment one can define multiple backends for one URL. But in Nextcloud one can only use one Backend. Is there a reason that you implemented multiple backends for one URL in the signaling config?

fancycode:
It's multiple backends for one hostname (not for one URL). A user could have two different Nextcloud installations at `https://cloud.domain.invalid/customer1/` and `https://cloud.domain.invalid/customer2/`.

src/signaling/backend_configuration_test.go
```diff
@@ -163,3 +164,166 @@ func TestIsUrlAllowed_AllowAll(t *testing.T) {
...
+	original_config.AddOption("backend1", "secret", string(testBackendSecret)+"-backend1")
+	original_config.AddOption("backend2", "url", "http://domain2.invalid")
+	original_config.AddOption("backend2", "secret", string(testBackendSecret)+"-backend2")
+	o_cfg, err := NewBackendConfiguration(original_config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	new_config := goconf.NewConfigFile()
+	new_config.AddOption("backend", "backends", "backend1, backend2")
```
fancycode, line 250: Please remove `backend2` from the list of ids.

src/signaling/backend_configuration_test.go
```diff
@@ -163,3 +164,166 @@ func TestIsUrlAllowed_AllowAll(t *testing.T) {
...
+
+	o_cfg.Reload(original_config)
+	if !reflect.DeepEqual(n_cfg, o_cfg) {
+		t.Error("BackendConfiguration should be equal after Reload")
+	}
+}
+
+func TestBackendReloadAddBackend(t *testing.T) {
+	original_config := goconf.NewConfigFile()
+	original_config.AddOption("backend", "backends", "backend1, backend2")
```
fancycode, line 270: Please remove `backend2` from the list of ids.

src/signaling/backend_configuration_test.go
```diff
@@ -163,3 +164,166 @@ func TestIsUrlAllowed_AllowAll(t *testing.T) {
...
+	original_config.AddOption("backend1", "secret", string(testBackendSecret)+"-backend1")
+	original_config.AddOption("backend2", "url", "http://domain1.invalid")
+	original_config.AddOption("backend2", "secret", string(testBackendSecret)+"-backend2")
+	o_cfg, err := NewBackendConfiguration(original_config)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	new_config := goconf.NewConfigFile()
+	new_config.AddOption("backend", "backends", "backend1, backend2")
```
fancycode, line 314: Please remove `backend2` from the list of ids.

src/signaling/backend_configuration_test.go
```diff
@@ -163,3 +164,166 @@ func TestIsUrlAllowed_AllowAll(t *testing.T) {
...
+	original_config.AddOption("backend2", "url", "http://domain2.invalid")
+	original_config.AddOption("backend2", "secret", string(testBackendSecret)+"-backend2")
+
+	o_cfg.Reload(original_config)
+	if !reflect.DeepEqual(n_cfg, o_cfg) {
+		t.Error("BackendConfiguration should be equal after Reload")
+	}
+}
+
+func TestBackendReloadRemove(t *testing.T) {
```
fancycode, line 300: Isn't this the same test as `TestBackendReloadRemoveBackend`?

src/signaling/backend_configuration.go
```diff
@@ -70,40 +72,15 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration,
...
 	} else if backendIds, _ := config.GetString("backend", "backends"); backendIds != "" {
-		seenIds := make(map[string]bool)
-		for _, id := range strings.Split(backendIds, ",") {
-			id = strings.TrimSpace(id)
-			if id == "" {
-				continue
+		for host, backendIDs := range getConfiguredHosts(config) {
+			for id, configuredBackends := range backendIDs {
+				backends[host] = append(backends[host], configuredBackends...)
+				var u strings.Builder
```
fancycode, line 78: With the map structure change below you can use the (single) backend directly, no need for `strings.Builder`.

fancycode (changes requested):
Thanks, looks much better now. Only a couple of minor changes remain until this can be merged.

jkahrs:
> Thanks, looks much better now. Only a couple of minor changes remain until this can be merged.

Thank you for the review. I will take a look at it and update the PR

fancycode:
Thanks!

PR #53, opened 2020-10-06
Minor reload improvements

fancycode:
Follow-up on #52 

- Always load backends in the order they are configured.
- Don't allow reload with old-style backend configurations.
- Update reload tests.
- Log what backends are added/changed/removed on reload.

cc @membero @jkahrs

### DIFF
--- a/src/signaling/backend_configuration.go
+++ b/src/signaling/backend_configuration.go
@@ -71,7 +71,7 @@ func NewBackendConfiguration(config *goconf.ConfigFile) (*BackendConfiguration, 
 			compat: true,
 		}
 	} else if backendIds, _ := config.GetString("backend", "backends"); backendIds != "" {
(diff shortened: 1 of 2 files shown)


src/signaling/backend_configuration.go
```diff
@@ -146,26 +146,28 @@ func (b *BackendConfiguration) UpsertHost(host string, backends []*Backend) {
...
 
-			ids[id] = true
+		if seen[id] {
+			continue
 		}
+		ids = append(ids, id)
+		seen[id] = true
 	}
 
 	return ids
```
jkahrs, line 179: Just a remark: This guarantees the ids to be in config order. Could it be useful to also sort the ids?

fancycode: I would rather have the backends being processed in the order they are configured in the file.

PR #57, opened 2020-10-28
Add '--full-trickle' to janus command

SystemKeeper:
The signaling server warns about the missing '--full-trickle' parameter, therefore adding this.


fancycode:
Thanks!

PR #82, opened 2021-02-18
fix(docs): update config example for Apache proxy config

schleiftier:
Without the trailing slash, the config will not work. Instead the Nextcloud talk client will receive an error 400.

See also https://help.nextcloud.com/t/error-400-when-trying-to-access-high-performance-backend-signaling/106999/2

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,7 +269,7 @@ interface on port `8080` below):
 
         RewriteEngine On
         # Websocket connections from the clients.
-        RewriteRule ^/standalone-signaling/spreed$ - [L]
+        RewriteRule ^/standalone-signaling/spreed/$ - [L]
         # Backend connections from Nextcloud.
         RewriteRule ^/standalone-signaling/api/(.*) http://127.0.0.1:8080/api/$1 [L,P]
 


README.md
```diff
@@ -269,7 +269,7 @@ interface on port `8080` below):
 
         RewriteEngine On
         # Websocket connections from the clients.
-        RewriteRule ^/standalone-signaling/spreed$ - [L]
+        RewriteRule ^/standalone-signaling/spreed/$ - [L]
```
fancycode, line 272: Thinking about this, now the rule is no longer applied to client requests (which are loading `/standalone-signaling/spreed` (without a trailing slash). Does it still work if you remove the line (and the comment above) completely?

schleiftier: Without that line I Nextcloud does not complain and I can join calls, but the video connection is not working and chat messages are not loaded or delivered.

schleiftier: Others seemed to have a similar problem with the nginx config as well: https://help.nextcloud.com/t/signaling-server-recommendations/52798/58?u=freed

Lebernd: My [post](https://help.nextcloud.com/t/error-400-when-trying-to-access-high-performance-backend-signaling/106999/6)in the forum. I can only recommend the commit.

fancycode:
Thanks and sorry this was pending for too long.

PR #97, opened 2021-04-18
[Fix][Documentation] Fix Caddy v2 Example Config

Blackclaws:
There is a small but significant error in the Caddy v2 example configuration. The reverse proxy directive applies to the new path not to the old path hence you need to also strip the prefix from the reverse proxy directive in order for it to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ myserver.domain.invalid {
 myserver.domain.invalid {
   route /standalone-signaling/* {
     uri strip_prefix /standalone-signaling
-    reverse_proxy /standalone-signaling/* http://127.0.0.1:8080
+    reverse_proxy /* http://127.0.0.1:8080
   }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -253,7 +253,7 @@ myserver.domain.invalid {
 myserver.domain.invalid {
   route /standalone-signaling/* {
     uri strip_prefix /standalone-signaling
-    reverse_proxy /standalone-signaling/* http://127.0.0.1:8080
+    reverse_proxy /* http://127.0.0.1:8080
```
fancycode, line 256: Can't this be simplified as `reverse_proxy http://127.0.0.1:8080`?

Blackclaws: Ah, yes it can. I'm not a Caddy expert by far but let me change that real quick

PR #127, opened 2021-06-25
Support explicitly pinning streams to Janus event loops.

fancycode:
Needs Janus v0.11.4


fancycode:
No longer needed with the multi-proxy support.

PR #132, opened 2021-07-07
Force HTTPS for backend connections in old-style configurations.

### LINKED ISSUE #130: Force HTTPS connections to Nextcloud instance
Right now a user could force a downgrade by setting a HTTP URL directly inside the `hello` message. To avoid any potential sensitive data being transmitted unencrypted we should consider always using HTTPS. (or requiring users to also configure the protocol for any given backend)

fancycode:
Prevents clients from sending a HTTP auth URL to trigger an insecure connection from the signaling server to Nextcloud.

@LukasReschke FYI

Fixes #130


LukasReschke (commented):
🎉

fancycode:
Thanks!

PR #134, opened 2021-07-07
Only include body in 307/308 redirects if going to same host.

fancycode:
@LukasReschke FYI


LukasReschke (approved):
Thanks! Seems reasonable :-)

fancycode:
Thanks

PR #157, opened 2021-09-28
Support separate permissions for publishing audio / video.

fancycode:
Clients with permission "publish-media" can publish any audio/video. Other clients need to have "publish-audio" and/or "publish-video" to publish the corresponding media types. For this the SDP in the offer is parsed to check what will be sent.

CC @nickvergessen


fancycode:
Do we need a way to notify Talk that the new permissions are available? I assume you want to sent the old permission if the signaling server is running an old version.

nickvergessen (approved):
Looks good

nickvergessen:
> Do we need a way to notify Talk that the new permissions are available? I assume you want to send the old permission if the signaling server is running an old version.

I'm more thinking of showing a related error in the admin UI and you need to use the latest version (which understands both) before updating to 23

fancycode:
Do you notify the signaling server about the changed permissions (with a `participants` request)? If not, the new permissions are only fetched when the user joins the room again.

fancycode:
Any other changes required?

nickvergessen:
Can you give be another compiled binary? 😎

nickvergessen:
Works, patch for version checking will come in https://github.com/nextcloud/spreed/pull/6442

So this here can go in

fancycode:
Great, thanks!

nickvergessen:
Could we also return the features on the hello response of the websocket?
Then I can show a user warning too

fancycode:
> Could we also return the features on the hello response of the websocket? Then I can show a user warning too

The features are already in there (and used to detect presence of Janus).

PR #169, opened 2021-11-08
Check individual audio/video permissions on change.

fancycode:
Follow-up to #157 to support individual permissions when a session is changed.

E.g. if a client publishes audio/video and no longer has the video permission, the whole publisher will be closed. Previously this was only checking the generic "media" permission.

CC @nickvergessen


nickvergessen (approved):
Seems to work still properly.
Currently the web does a force reconnect IIRC, so maybe we could avoid this now/in the future? @danxuliu

PR #224, opened 2022-04-06
Migrate to structured logging using "go.uber.org/zap".

fancycode:


fancycode:
Go supports native structured logging through the [slog](https://pkg.go.dev/log/slog) package since 1.21, so this should be used instead of `go.uber.org/zap`.

For 1.20, the `https://pkg.go.dev/golang.org/x/exp/slog` could be used.

Issue #246, opened 2022-05-06
Combine session updates of multiple rooms into 1 request


nickvergessen:
As discussed.

Ticket with info: https://github.com/nextcloud/spreed/issues/7335
Draft implementation: https://github.com/nextcloud/spreed/pull/7338

Config can be enabled with:
```
occ config:app:set spreed "session-ping-limit" --value="200"
```

Capability:
```json
        "spreed": {
          "features": [
            "audio",
            "video",
            "chat-v2",
            "conversation-v4",
…
          ],
          "config": {
            …
            "signaling": {
              "session-ping-limit": 0
            }
          }
        },
```

Open questions: Currently we return the room token in the response:
https://github.com/nextcloud/spreed/pull/7338/files#diff-9fc0995666c1b2714b4c68a64ca745e909c57584e059f40e3aec6ba76e10e408L763

I guess you wouldn't expect that anymore as you also don't send it anymore?

fancycode:
> I guess you wouldn't expect that anymore as you also don't send it anymore?

Yeah, the response data is currently ignored anyway:
https://github.com/strukturag/nextcloud-spreed-signaling/blob/cbcd27bd74c9a5c13846a11d95fb17e1d7bfcbb9/room.go#L833-L836

PR #250, opened 2022-05-12
Combine ping requests of different rooms on the same backend.

fancycode:
- [x] Add tests

Requires https://github.com/nextcloud/spreed/pull/7338

Implements #246

### DIFF
--- a/hub.go
+++ b/hub.go
@@ -127,6 +127,7 @@ type Hub struct {
 	rooms    map[string]*Room
 
 	roomSessions    RoomSessions
+	roomPing        *RoomPing
 	virtualSessions map[string]uint64
 
 	decodeCaches []*LruCache
@@ -210,6 +211,11 @@ func NewHub(config *goconf.ConfigFile, nats NatsClient, r *mux.Router, version s
 	}
 
 	roomSessions, err := NewBuiltinRoomSessions()
+	if err != nil {
+		return nil, err
+	}
+
+	roomPing, err := NewRoomPing(backend, backend.capabilities)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


fancycode:
@nickvergessen FYI

nickvergessen:
- [ ] Add feature flag header so the Nextcloud part can show a "upgrade needed" error

fancycode:
> * [x]  Add feature flag header so the Nextcloud part can show a "upgrade needed" error

Added as `multiroom-pings`.

fancycode:
@nickvergessen this is ready for testing against your Talk changes

nickvergessen (approved):
Works, only need to adjust the log on our side:
`Pinged 2 sessions in room   `

Also set the value to 1 session per request to test the chunking:
```
│[debug] 2022-05-17T18:18:59.000 spreed-hpb       Pinged 1 sessions in room N/A                                                                                         
│[debug] 2022-05-17T18:18:59.000 spreed-hpb       Pinged 1 sessions in room N/A                                                                                         
│[debug] 2022-05-17T18:19:09.000 spreed-hpb       Pinged 1 sessions in room N/A                                                                                        
│[debug] 2022-05-17T18:19:09.000 spreed-hpb       Pinged 1 sessions in room N/A    
```

So works greatly

nickvergessen:
But while thinking on the deployment order, we can both release the update immediately, right?
On our side we lessen the strictness and publish a capability and you only do the multi request when it is there.
So we don't need to tell people to update on Nextcloud.

So we could even backport it on our side and if people update the HPB, they would automatically benefit from the change right away?

fancycode:
> But while thinking on the deployment order, we can both release the update immediately, right? On our side we lessen the strictness and publish a capability and you only do the multi request when it is there. So we don't need to tell people to update on Nextcloud.
> 
> So we could even backport it on our side and if people update the HPB, they would automatically benefit from the change right away?

I think we could release independently: the old signaling server will work as before and still work with the changes in Talk. With this PR merged, the batched pings will only be sent if the `session-ping-limit` value is there (i.e. with a new version of Talk). The old server with a new version of Talk will also work as before.
There is no need for a capability, the check for `session-ping-limit` should be sufficient.

fancycode:
> If you want to have a shorter list you could remove it again as we don't seem to need it.

Ah right, true.

PR #261, opened 2022-05-16
Update Alpine to 3.15 version, fix CVE-2022-28391

solracsf:
See https://github.com/advisories/GHSA-h8c3-8522-vxc6


fancycode:
Thanks!

PR #314, opened 2022-08-05
Official docker images.

fancycode:


alexanderadam:
This looks perfect to me :star_struck:

fancycode:
> This looks perfect to me star_struck

Thanks, first images are avaliable here https://hub.docker.com/r/strukturag/nextcloud-spreed-signaling

jakobroehrl:
Thanks a lot. @fancycode 
Do I still need Janus, Nuts, nginx? Or is all of it packed into the docker file?

fancycode:
Our packages only provide the signaling server / -proxy. You will need to run the other services (either manually or from other images), so they can be used by this ones.

Issue #315, opened 2022-08-07
Ubuntu20.04 : failure after package update


francoisPE:
Hello,
I have a ubuntu VPS with a signaling server
My configuration was working properly up to yesterday when I upgrade package... (to version 1.0.2)
I see almost no message in syslog
```Aug  8 00:21:06 ns2 systemd[1]: signaling.service: Scheduled restart job, restart counter is at 170.
Aug  8 00:21:06 ns2 systemd[1]: Stopped Nextcloud Talk signaling server.
Aug  8 00:21:06 ns2 systemd[1]: Started Nextcloud Talk signaling server.
Aug  8 00:21:06 ns2 systemd[1]: signaling.service: Main process exited, code=killed, status=31/SYS
Aug  8 00:21:06 ns2 systemd[1]: signaling.service: Failed with result 'signal'.
Aug  8 00:21:06 ns2 kernel: [8136072.888627] audit: type=1326 audit(1659910866.720:18766): auid=4294967295 uid=126 gid=134 ses=4294967295 pid=2544650 comm="signaling" exe="/usr/bin/signaling" sig=31 arch=c000003e syscall=160 compat=0 ip=0x40444e code=0x80000000
Aug  8 00:21:08 ns2 systemd[1]: Stopped Nextcloud Talk signaling server.
```
service signaling status says
```● signaling.service - Nextcloud Talk signaling server
     Loaded: loaded (/lib/systemd/system/signaling.service; enabled; vendor preset: enabled)
     Active: activating (auto-restart) (Result: signal) since Mon 2022-08-08 00:31:49 CEST; 195ms ago
    Process: 2547319 ExecStart=/usr/bin/signaling --config /etc/signaling/server.conf (code=killed, signal=SYS)
   Main PID: 2547319 (code=killed, signal=SYS)

Aug 08 00:31:49 ns2.dgtz.fr systemd[1]: signaling.service: Main process exited, code=killed, status=31/SYS
Aug 08 00:31:49 ns2.dgtz.fr systemd[1]: signaling.service: Failed with result 'signal'.
```
That is very short...
Is there any incompatibility to check ?

fancycode:
There is no version 1.0.2 of the signaling server yet, so I'm wondering what you are running. Was this prebuilt or did you compile it manually?

`SIGSYS` according to the manpage is `Bad system call (SVr4)`, so maybe you are enforcing some kind of security rules which block operations of the signaling server?

[...]

PR #322, opened 2022-08-19
remove @resources from SystemCallFilter

morph027:
Introduced in https://github.com/strukturag/nextcloud-spreed-signaling/pull/276.

* https://gitlab.com/packaging/nextcloud-spreed-signaling/-/commit/3fcaf45b6f32276ad1c42d0ff6a337db4d66da79#note_1055183029
* https://github.com/nextcloud/vm/issues/2371
* https://github.com/strukturag/nextcloud-spreed-signaling/issues/315

As @Tachi107 left a comment about `ProcSubset`, i tested the other way around and leave ´@resources` in place while removing `ProcSubset=pid`, still does not work.

### DIFF
--- a/dist/init/systemd/signaling.service
+++ b/dist/init/systemd/signaling.service
@@ -37,7 +37,7 @@ RestrictRealtime=yes
 RestrictSUIDSGID=yes
 SystemCallArchitectures=native
 SystemCallFilter=@system-service
-SystemCallFilter=~ @privileged @resources
+SystemCallFilter=~ @privileged
 
 [Install]
 WantedBy=multi-user.target


Tachi107 (approved):
Hey @morph027, thanks for tracking down this issue!

I can't say I'm surprised, as the hardening measures were on the strict side, but I'm happy to know that everything else doesn't seem to create issues.

LGTM!

Tachi107:
It's a bit odd that I didn't find this issue in my setup, but maybe the usage of some new syscall has been introduced in the latest release. Sorry for the issues I caused :)

morph027:
NP ;) Better safe than sorry in terms of security.

fancycode:
Thanks!

PR #428, opened 2023-02-22
TURN_API_KEY and TURN_SECRET fix

3x3cut0r:
the docker environment variables TURN_API_KEY and TURN_SECRET didn't work this is because the entrypoint.sh is only replacing the values but do not uncomment the line like the other replacements do


fancycode:
Thanks!

PR #429, opened 2023-02-22
Fix example in docker README

SystemKeeper:
The parameter should be `BACKEND_<NAME>_SHARED_SECRET` and not `BACKEND_<NAME>_SECRET`.


fancycode:
Thanks!

PR #430, opened 2023-02-23
Allow SKIP_VERIFY in docker image

SystemKeeper:
To allow using the official docker image of the signaling server for development as well, it would be nice to add `SKIP_VERIFY` as an environment option as well. 

If this was purposely not added, because it's considered a development option, feel free to just close this.


fancycode:
Thanks!

PR #432, opened 2023-02-23
TURN_API_KEY  and TURN_SECRET fix

3x3cut0r:
sry , my last fix wasn't consistent to the other replacements now we can update not only the default value of TURN_API_KEY and TURN_SECRET

some note:
maybe someone can change the TURN_SECRET value "secret" in something like "turnsecret" to not run into trouble with other "secret" value(s) in the backend section.


fancycode:
Replaced by #439

PR #439, opened 2023-03-06
docker: Don't rely on default values when updating TURN settings.

fancycode:
Replaces #432

@3x3cut0r Please test, this is better than your change as it won't replace other `secret`  values in the config files.

### DIFF
--- a/docker/server/entrypoint.sh
+++ b/docker/server/entrypoint.sh
@@ -129,13 +129,13 @@ if [ ! -f "$CONFIG" ]; then
 
   if [ ! -z "$SKIP_VERIFY" ]; then
     sed -i "s|#skipverify =.*|skipverify = $SKIP_VERIFY|" "$CONFIG"
-  fi 
+  fi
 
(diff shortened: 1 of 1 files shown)


fancycode:
> i cannot test it atm but looking at the code ... why do you escape and insert the \? in the search pattern?: `#\?apikey` `#\?secret`

This is so both `apikey = foo`  and `#apikey = foo`  get updated. Same for the TURN `secret`.

3x3cut0r:
i understand the '?' inside the sed command and i know what you are trying to do, but is the backslach '\' infront of the questionmark not escaping this function of 'the char before is optional' to a simple character '?' ?

fancycode:
Sed requires "zero-or-no-matches" to be written as `\?`:
https://www.gnu.org/software/sed/manual/html_node/Regular-Expressions.html

3x3cut0r:
interesting ... i never ever did it like this.
then i should be fine!
then maybe it should also be done like this on the other vars, no?

Issue #79, opened 2021-02-10
Add the ability to configure supported codecs and profiles


Ivansss:
There should be a way to configure supported codecs and profiles.
At the moment Janus forces VP8 as the only video codec to use:
https://github.com/meetecho/janus-gateway/blob/master/plugins/janus_videoroom.c#L2306

Gymnae:
Created an upstream issue, which might help: 
https://github.com/meetecho/janus-gateway/issues/3205

PR #469, opened 2023-04-26
Offer more video codecs for Janus rooms

Gymnae:
Looking at this issue here: https://github.com/strukturag/nextcloud-spreed-signaling/issues/79
and the example config of Janus https://github.com/meetecho/janus-gateway/blob/master/conf/janus.plugin.videoroom.jcfg.sample I made these small changes.
Works locally for me with my docker compose setup.

Note: I am using a newer version of Janus, v1.1.3

### DIFF
--- a/mcu_janus.go
+++ b/mcu_janus.go
@@ -734,10 +734,13 @@ func (m *mcuJanus) getOrCreatePublisherHandle(ctx context.Context, id string, st
 		"publishers": 1,
 		// Do not use the video-orientation RTP extension as it breaks video
 		// orientation changes in Firefox.
-		"videoorient_ext": false,
-		// Offer more video & audio codecs
-		"audiocodec": "opus,g722,pcmu,pcma,isac32",
-		"videocodec": "vp9,h265,vp8,h264,av1",
+                // "videoorient_ext": false,
+                // Offer more video & audio codecs
+                "audiocodec": "opus,g722,pcmu,pcma,isac32",
+                "videocodec": "av1,vp9,vp8,h265,h264",
+                "video_svc": true,
+                "vp9_profile": "2",
+                "h264_profile": "42e01f",
 	}
 	var maxBitrate int
 	if streamType == streamTypeScreen {
--- a/mcu_janus.go
+++ b/mcu_janus.go
@@ -738,7 +738,6 @@ func (m *mcuJanus) getOrCreatePublisherHandle(ctx context.Context, id string, st
 		// Offer more video & audio codecs
 		"audiocodec": "opus,g722,pcmu,pcma,isac32",
 		"videocodec": "vp9,h264,h265,vp8,av1",
-        }
 	}
 	var maxBitrate int
 	if streamType == streamTypeScreen {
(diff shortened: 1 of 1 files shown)


fancycode:
Thanks for your PR. First of all, the lists should be configurable and empty by default to not change the existing behaviour.

Which codecs are used with these changes? Did you test between different browsers and the mobile apps (iOS / Android)?

Gymnae:
> Thanks for your PR. First of all, the lists should be configurable and empty by default to not change the existing behaviour.

I agree, but I also have to confess: I'm no coder, I don't understand your code well enough to add this feature, so I chose a set of codecs and an order which made sense to me. 

> Which codecs are used with these changes? 

I can confirm that VP9 and h265 appear to be used, at least the built-in tools for WebRTC check tell me that

> Did you test between different browsers and the mobile apps (iOS / Android)?

I tested:

| Client 1             | Client 2               | Result  |
|----------------------|------------------------|---------|
| macOS Chrome         | macOS Chrome           | &check; |
| macOS Safari         | macOS Chrome           | &check; |
| macOS Firefox        | macOS Firefox          | &check; |
| Linux (KDE) Chrome   | macOS Chrome           | &check; |
| Linux (KDE) Chrome   | Windows Chrome         | &check; |
| Windows Chrome       | macOS Chrome           | &check; |
| Windows Firefox      | macOS Firefox          | &check; |
| Windows Edge         | macOS Chrome           | &check; |
| all Desktop Browsers | Talk iOS               | &cross; |
| all Desktop Browsers | mobile Safari (iOS 16) | &check; |
| Talk iOS             | Talk iOS               | &check; |

fancycode:
> *The issue with Talk iOS might be related to supported codecs, I raised an issue regarding this: [nextcloud/talk-ios#1198](https://github.com/nextcloud/talk-ios/issues/1198) Audio works, but no incoming video is shown, just the default background colour without the user icon – I guess the app sees a video stream, but can't deal with it

So this will need to be fixed first to not break existing installations.

Gymnae:
> > *The issue with Talk iOS might be related to supported codecs, I raised an issue regarding this: [nextcloud/talk-ios#1198](https://github.com/nextcloud/talk-ios/issues/1198) Audio works, but no incoming video is shown, just the default background colour without the user icon – I guess the app sees a video stream, but can't deal with it
> 
> So this will need to be fixed first to not break existing installations.

Agreed. Hopefully it’s not a complicated fix, seeing that mobile Safari supports the codecs.

Gymnae:
@fancycode – After experimenting more with the list `videocodec` it occurred to me that the issue might not lie with Nextcloud Talk clients needing to support more codecs. The codecs are coming from integrated libraries, the iOS app, for example, supports AV1, but not VP9. Firefox plays well with VP9, but not AV1. Chromium does everything.

Since the list `videocodec` is just a list of offered codecs for negotiation, the issue may lie with this signalling backend. The server appears not setup for renegotiation between subscribers and publisher. Why would it be, if by default only VP8 is enabled. Currently, clients choose their preferred codec from the list and stick with it, even if another subscriber "enters the room" not supporting the current codec. I might not fully grasp the difference between publisher and subscriber here, but it appears that there's no acting upon SDP messages including offered candidates.

I can see that required SDP messages are exchanged between the peers, but the codec is not re-negotiated. 
Is that something your backend could offer? Then the `videocodec` list could remain hardcoded and peers would renegotiate.

PR #487, opened 2023-06-11
Write backends comma-separated to config

### LINKED ISSUE #486: Docker image with multiple backends not working
I have the following environment variables set in a docker compose file:

```
      BACKENDS: "nextcloud nextcloud2"
      BACKEND_NEXTCLOUD_URL: "http://nextcloud.local"
      BACKEND_NEXTCLOUD_SHARED_SECRET: "44444444444444444444444444444444"
      BACKEND_NEXTCLOUD2_URL: "http://nextcloud2.local"
      BACKEND_NEXTCLOUD2_SHARED_SECRET: "44444444444444444444444444444444"
```

According to the docs, the backends should be space separated:
https://github.com/strukturag/nextcloud-spreed-signaling/blob/5a84dd7d656670ec80e9c32847f1db1980443145/docker/README.md?plain=1#L24

Example from the docs:
https://github.com/strukturag/nextcloud-spreed-signaling/blob/5a84dd7d656670ec80e9c32847f1db1980443145/docker/README.md?plain=1#L69-L81

This results in the following error when starting the container:

> backend_storage_static.go:200: Backend nextcloud nextcloud2 is missing or incomplete, skipping

According to the server config, the backends property should be comma-separated:
https://github.com/strukturag/nextcloud-spreed-signaling/blob/5a84dd7d656670ec80e9c32847f1db1980443145/server.conf.in#L61-L67

but the above docker command results in a config file being written as

> backends = nextcloud nextcloud2

SystemKeeper:
Fixes #486

### DIFF
--- a/docker/server/entrypoint.sh
+++ b/docker/server/entrypoint.sh
@@ -216,7 +216,8 @@ if [ ! -f "$CONFIG" ]; then
   fi
 
   if [ ! -z "$BACKENDS" ]; then
-    sed -i "s|#backends = .*|backends = $BACKENDS|" "$CONFIG"
+    BACKENDS_CONFIG=$(echo $BACKENDS | sed 's/ /,/g')
+    sed -i "s|#backends = .*|backends = $BACKENDS_CONFIG|" "$CONFIG"
 
     echo >> "$CONFIG"
     for backend in $BACKENDS; do


fancycode:
Thanks! I might need to add tests for the Docker entrypoints at some point...

PR #488, opened 2023-06-11
Add allowall to docker image

SystemKeeper:
For easier work in a development environment (e.g. https://github.com/juliushaertl/nextcloud-docker-dev) it would be nice if we could set the `allowall` property directly in the docker image.


fancycode:
Thanks

PR #602, opened 2023-11-15
Include "~docker" in version if built on Docker.

### LINKED ISSUE #601: Include version on docker image
It would be nice if the docker image could also report the version it was build from instead of the commit hash. For support reasons I would suggest to have it differ slightly from the HPB repo, so instead of displaying „1.1.2“ something like „1.1.2-docker“ (or the like) would be nice.

fancycode:
Resolves #601 

CC @SystemKeeper

### DIFF
--- a/docker/proxy/Dockerfile
+++ b/docker/proxy/Dockerfile
@@ -5,7 +5,8 @@ ARG TARGETOS
 WORKDIR /workdir
 
 COPY . .
-RUN apk add --no-cache bash git build-base protobuf && \
+RUN touch /.dockerenv && \
+    apk add --no-cache bash git build-base protobuf && \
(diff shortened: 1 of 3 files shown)


fancycode:
Unfortunately the Docker environment is not detected correctly on the CI builders (which build the official images). Needs more investigation...

SystemKeeper:
You can check for CI environment: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

fancycode:
> You can check for CI environment variable: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

Yeah, however I'm not sure these environments are passed through to the Docker build environment. Also I don't want to unconditionally use `~docker` in the version if built on CI.

SystemKeeper:
It ain’t stupid if it works 😂

PR #750, opened 2024-05-27
Throttle /64 subnets for IPv6.

fancycode:
Also addresses comments from #747.

@danxuliu FYI


danxuliu (approved):
Changes look good :+1: 

Thanks!

PR #765, opened 2024-06-19
Don't throttle valid but expired resume requests.

fancycode:
@SystemKeeper FYI


SystemKeeper (approved):
Tested and works fine. Reverted the changes in https://github.com/nextcloud/talk-ios/pull/1676 for the test.

PR #780, opened 2024-07-24
Improve caching when fetching capabilities.

fancycode:
- Honor "max-age" in "Cache-Control" header.
- Use "ETag" to avoid fetching unmodified data.

@nickvergessen I could not find where it is set, but Nextcloud seems to always include an "cache-control: max-age=60" header. I'd rather use the caching headers as set by the server instead of hardcoding a cache time of 1 hour.


nickvergessen:
> I could not find where it is set, but Nextcloud seems to always include an "cache-control: max-age=60" header.

I think this the fallback via htaccess/nginx config?
Don't see it in PHP code

fancycode:
> I think this the fallback via htaccess/nginx config? Don't see it in PHP code

Thanks, that explains why I could not find it.

Looks like there was a suggestion to set a default cache policy `expires 1m` in the past that got removed in https://github.com/nextcloud/documentation/pull/5857 (which I have locally set in my nginx).

So I will assume none will be sent by default in recent setups. Should I just honor that and check for the capabilities whenever needed (using `If-None-Match` and the `ETag`) or should I forcefully cache the capabilities for some time?

nickvergessen:
Either way is fine I guess.

fancycode:
Ok, so I'll honor the `Cache-Control` header if one is set, otherwise I will use a default cache duration of one minute to avoid overloading a Nextcloud backend if many participants join at the same time.

PR #783, opened 2024-07-29
Enforce a minimum duration to cache capabilities.

fancycode:
Simply using the values from the "Cache-Control" header is problematic as recent setups of Nextcloud are not setting the header and capabilities would not be cached at all in these cases.

Follow-up to #780 

@SystemKeeper PTAL


SystemKeeper (approved):
Tested and works fine! Thanks!

fancycode:
Thanks for testing!

PR #792, opened 2024-07-31
Expose backend session limits through prometheus stats.

### LINKED ISSUE #787: Add sessionlimit to prometheus export
I think it would be useful to have the value of sessionlimit present in the Prometheus export.

Both the global value and the backend-specific values should be there. This would help in monitoring usage by backends and put in place an alerting system.

While these values can also be obtained with parsing the configuration file and/or the etcd database, to have them in the Prometheus export would make the process simpler.

fancycode:
Resolves #787 

@pmarini-nc FYI, don't know if you have the possibility to test this from source

### DIFF
--- a/backend_configuration_stats_prometheus.go
+++ b/backend_configuration_stats_prometheus.go
@@ -26,6 +26,12 @@ import (
 )
 
 var (
+	statsBackendLimit = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "signaling",
+		Subsystem: "backend",
+		Name:      "session_limit",
+		Help:      "The session limit of a backend",
+	}, []string{"backend"})
 	statsBackendLimitExceededTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "signaling",
 		Subsystem: "backend",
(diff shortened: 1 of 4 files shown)


pmarini-nc:
@fancycode , just tested it and it works fine. Thanks a lot!
Only thing maybe it is worth adding the global limit as well?

fancycode:
> Only thing maybe it is worth adding the global limit as well?

Not sure what you mean with "global limit". If the signaling server is running with compatibility settings (i.e. not with explicit backend definitions), the session limit will be available in the `compat` backend through the stats.

pmarini-nc:
Hello,

My understanding until now was the following: if a backend defines it, it gets its value of `sessionlimit`. If not it gets the value of a global sessionlimit, as defined in section  [backend].

However this seems not the case, that is: there is no fallback `sessionlimit` (like there is one for `secret`). If so, when we don't specify a sessionlimit in the backend, there is no limit, correct?

fancycode:
> However this seems not the case, that is: there is no fallback `sessionlimit` (like there is one for `secret`). If so, when we don't specify a `sessionlimit` in the backend, there is no limit, correct?

Right, adding a global sessionlimit as fallback for backends that have none defined is tracked in #796

PR #808, opened 2024-08-28
Send ping requests to local instance for federated sessions.

fancycode:
Fixes https://github.com/nextcloud/spreed/issues/12917

cc @nickvergessen


nickvergessen (approved):
Tested successfully now

Issue #845, opened 2024-10-22
`make build`: Nothing to be done for build


hanserasmus:
Hi all

The last time I updated my server was around June more or less. Today I did a `git pull` & a `make clean && make build`. My git pull did indeed pull a whole bunch of new files, but make build does not work. Am I being stupid as usual?

```
git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
```
```
/bin/signaling --version
nextcloud-spreed-signaling version f6125dac3ffa1e5cb5df57421293c95689ac6b57/go1.21.5
```

Thanks in advance for any help!
Cheers

fancycode:
Strange, could you please try removing the previously built files from the `bin` folder? (Asking myself why this is not done by `make clean`.)

If it still doesn't work after that, please provide the output of `make build --debug`. Thanks!

hanserasmus:
I checked earlier and there was only a ‘proxies’ file in there. Nothing else. Will do the debug tomorrow and report back. Sent from my iPhoneOn 22 Oct 2024, at 21:48, Joachim Bauch ***@***.***> wrote:﻿
Strange, could you please try removing the previously built files from the bin folder? (Asking myself why this is not done by make clean.)
If it still doesn't work after that, please provide the output of make build --debug. Thanks!

hanserasmus:
***@***.***:~# cd /opt/nextcloud-spreed-signaling/
***@***.***:/opt/nextcloud-spreed-signaling# make build --debug
GNU Make 4.2.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html
>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Reading makefiles...
Updating makefiles....
Updating goal targets....
 File 'build' does not exist.
Must remake target 'build'.
Successfully remade target file 'build'.
make: Nothing to be done for 'build'.
***@***.***:/opt/nextcloud-spreed-signaling#


Does this help?

fancycode:
I could reproduce it by setting an really old timestamp for my `bin` folder (e.g. sometime in 2023), then I get the same result. Executing `touch bin` updates the timestamp and `make build` works for me afterwards.

Can you confirm this works for you, too?

hanserasmus:
[...]

PR #848, opened 2024-10-28
make: Improve dependency tracking.

fancycode:
This should fix issues like #845 

@hanserasmus FYI


hanserasmus:
Nice! Thank you!

PR #868, opened 2024-11-19
Support relaying of chat messages.

fancycode:
@nickvergessen FYI


fancycode:
@miaulalala rebased and added test for federation

PR #871, opened 2024-11-24
Update docker-compose.yml

IamHaddad:
Fixed command and config file location


fancycode:
A better solution that doesn't use the filename also as path is in #881

PR #873, opened 2024-12-03
docker: apply Docker image versioning

solracsf:
- Remove deprecated `version` param
- Upgrade `nats` to latest stable (personally, I've always upgraded it on my deployement without any issues for 3y)
- Stick `coturn` to latest stable instead of `latest`


fancycode:
Thanks!

PR #911, opened 2025-01-28
Do not log nats url credentials

Adphi:


natsclient.go
```diff
@@ -153,3 +154,12 @@ func (c *natsClient) Decode(msg *nats.Msg, vPtr interface{}) (err error) {
 	}
 	return
 }
+
+func removeURLCredentials(u string) string {
+	url, err := url.Parse(u)
+	if err != nil {
+		return u
+	}
+	url.User = nil
```
fancycode, line 163: Imho it would be better to replace any existing `User` credentials with something like
```golang
url.UserPassword("***", "***")
```

That way, the log shows that credentials have been used but they are not visible.

fancycode:
Thanks!

PR #916, opened 2025-02-06
Migrate cache-control parsing to https://github.com/pquerna/cachecontrol

### LINKED ISSUE #914: github.com/marcw/cachecontrol - dependency unmaintained for 11yrs
@fancycode I have just been looking into updating the nextcloud-signaling-spreed package in Debian testing/unstable for Debian 13 and stumbled over the following problem:

Version 2.0.2 of this project is now (change with some release since 1.3.2) dependent on github.com/marcw/cachecontrol, a Golang project that hasn't seen any activity for the last 11 yrs. Please fork that project under your namespace (and continue its maintenance) or possibly depend to https://github.com/pquerna/cachecontrol instead (don't know whether that is doable, but that one is available in Debian and actively maintained).

I cannot bring an 11yrs-old project to Debian without having an upstream that maintains the code. And the signaling server should not use this code either, maybe, unless the future maintenance question is settled.

I'd be happy to get some feedback on how to proceed with this issue in Debian. Thanks.

fancycode:
The previously used code was unmaintained for a very long time and caused problems for packagers.

Fixes #914

### DIFF
--- a/capabilities.go
+++ b/capabilities.go
@@ -33,7 +33,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/marcw/cachecontrol"
+	"github.com/pquerna/cachecontrol/cacheobject"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


fancycode:
@sunweaver feel free to use this PR as patch for your packaging. Will be part of the next release (no timeframe yet).

PR #937, opened 2025-03-05
Add serverinfo API

fancycode:
@nickvergessen please take a look if this contains all information you want to show in the admin UI.

### DIFF
--- a/docs/standalone-signaling-api-v1.md
+++ b/docs/standalone-signaling-api-v1.md
@@ -1538,6 +1538,29 @@ for streams on the proxy. Only present if a bandwidth limit is configured on
 the proxy.
 
 
+### Dialout session
+
+If a SIP bridge with support for dial-out is connected, the serverinfo response
+will contain an additional property `dialout` with the following contents:
+
+    [
+      {
+        "sessionid": "the-session-id",
+        "connected": true,
+        "address": "192.168.1.0",
+        "useragent": "spreed-webrtc-sip-bridge/1.2.3",
+        "features": [
+          "start-dialout",
+          "datachannels",
+          "encryption"
+        ]
+      }
+    ]
+
+If the connection between SIP bridge and signaling server is interrupted, the
+`connected` property will be `false` and details on the session (`address`, `useragent` and `features`) omitted.
+
+
 ## Rooms API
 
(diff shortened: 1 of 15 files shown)


docs/standalone-signaling-api-v1.md
```diff
@@ -1416,6 +1416,151 @@ The signaling server provides an internal API that can be called from Nextcloud
...
+
+If a SIP bridge with support for dial-out is connected, the serverinfo response
+will contain an additional property `dialout` with the following contents:
+
+    [
+      {
+        "sessionid": "the-session-id",
+        "connected": true,
+        "address": "192.168.1.0",
+        "useragent": "spreed-webrtc-sip-bridge/1.2.3",
```
nickvergessen, line 1551: string split on / for version?
Or can we add a dedicated field?

fancycode: There is no version information that could be extracted from the `hello` request, so I will extract the version from the useragent and provide it as separate field.

nickvergessen: Thanks

fancycode:
@danxuliu I would like to return information about recording servers here too, if possible. Do they establish a permanent connection to the signaling server, or only when they are doing active recordings?

danxuliu:
> @danxuliu I would like to return information about recording servers here too, if possible. Do they establish a permanent connection to the signaling server, or only when they are doing active recordings?

Only when they are doing active recordings (they launch a browser and create an internal client of the signaling server in the browser).

What would you need from the recording server to show the information here? And what information would you like to show?

fancycode:
> Only when they are doing active recordings (they launch a browser and create an internal client of the signaling server in the browser).

Thanks, I see.

> What would you need from the recording server to show the information here? And what information would you like to show?

The idea was to have a way for the Talk admin UI to query information about the different backend components that might not be accessible directly (e.g. the signaling proxies or the SIP bridge).
I was thinking we could also at least provide the version of the recording backend, so the UI could show a hint about required updates if the running version is too old.

nickvergessen:
The recording backend has a reachable welcome endpoint that we already read:
Bildschirmfoto vom 2025-03-07 10-57-55

PR #965, opened 2025-04-11
Allow to set GOMAXPROCS

zc-devs:
Currently, I cannot override GOMAXPROCS, because it is hardcoded to be as much as CPUs number.
Since Go 1.5 the default value already set as number of CPUs => previous logic applies by default.
Patch allows to set/override default behavior via environment variable.

https://stackoverflow.com/questions/17853831/what-is-the-gomaxprocs-default-value
https://pkg.go.dev/runtime#GOMAXPROCS

### DIFF
--- a/client/main.go
+++ b/client/main.go
@@ -517,9 +517,7 @@ func main() {
 	}
 	cookie := signaling.NewSessionIdCodec([]byte(hashKey), blockBytes)
 
-	cpus := runtime.NumCPU()
-	runtime.GOMAXPROCS(cpus)
-	log.Printf("Using a maximum of %d CPUs", cpus)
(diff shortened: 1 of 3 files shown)


fancycode:
Nice find! Could you please also change this for the proxy and client codes?

https://github.com/strukturag/nextcloud-spreed-signaling/blob/6369cc8f3e0e2cc800c3476f74567bf2b231fd76/proxy/main.go#L79-L81
https://github.com/strukturag/nextcloud-spreed-signaling/blob/6369cc8f3e0e2cc800c3476f74567bf2b231fd76/client/main.go#L520-L522

Thanks!

fancycode:
Thanks for your contribution.

PR #973, opened 2025-04-16
Add metrics for backend client requests.

### LINKED ISSUE #968: Response time / violation as prometheus stat
It would be good to be able to see 2 more stats in the prometheus API:
- Response times (or average)
- Number of requests that ran into the "response timeout" in the last X

The idea would be to make it visible when a call had issues that it could be caused by this ad have it easily trackable via monitoring.

fancycode:
Resolves #968

@nickvergessen FYI

### DIFF
--- a/backend_client.go
+++ b/backend_client.go
@@ -30,6 +30,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/dlintw/goconf"
 )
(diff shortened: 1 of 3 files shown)


backend_client.go
```diff
@@ -117,11 +122,13 @@ func (b *BackendClient) PerformJSONRequest(ctx context.Context, u *url.URL, requ
 		return fmt.Errorf("no url passed to perform JSON request %+v", request)
 	}
 
-	secret := b.backends.GetSecret(u)
-	if secret == nil {
-		return fmt.Errorf("no backend secret configured for for %s", u)
+	backend := b.backends.GetBackend(u)
+	if backend == nil {
+		return fmt.Errorf("no backend configured for for %s", u)
```
nickvergessen, line 127:
```suggestion
		return fmt.Errorf("no backend configured for %s", u)
```

nickvergessen (approved):
Nice :)

PR #1112, opened 2025-11-13
Also rewrite token in comment for federated chat relay.

fancycode:
@miaulalala PTAL


Antreesy (approved):
Tested with local federations. Thanks!

PR #1175, opened 2026-01-14
Don't need all users in requests from Talk

### LINKED ISSUE #1173: Get rid of list of all users in requests from Talk to the signaling server
- Request when inviting / disinviting a user to a meeting (used to notify other users that know about that room but might be connected to another room):
https://github.com/strukturag/nextcloud-spreed-signaling/blob/d6264dfe1dbe79d3b8551feda51a0dd245ea3db2/talk/api.go#L116
https://github.com/strukturag/nextcloud-spreed-signaling/blob/d6264dfe1dbe79d3b8551feda51a0dd245ea3db2/talk/api.go#L125

- Request when a room is updated (used to notify other users that know about that room but might be connected to another room):
https://github.com/strukturag/nextcloud-spreed-signaling/blob/d6264dfe1dbe79d3b8551feda51a0dd245ea3db2/talk/api.go#L129

- Request when a room is deleted (used to notify other users that know about that room but might be connected to another room):
https://github.com/strukturag/nextcloud-spreed-signaling/blob/d6264dfe1dbe79d3b8551feda51a0dd245ea3db2/talk/api.go#L135

- Request when the "InCall" flag of a user changed:
https://github.com/strukturag/nextcloud-spreed-signaling/blob/d6264dfe1dbe79d3b8551feda51a0dd245ea3db2/talk/api.go#L143

- Request when user properties changed:
https://github.com/strukturag/nextcloud-spreed-signaling/blob/d6264dfe1dbe79d3b8551feda51a0dd245ea3db2/talk/api.go#L148

- Request when switching sessions to a different room (might not be relevant for large rooms):
https://github.com/strukturag/nextcloud-spreed-signaling/blob/d6264dfe1dbe79d3b8551feda51a0dd245ea3db2/talk/api.go#L165-L170

@danxuliu FYI

fancycode:
- [x] BackendRoomInviteRequest
- [x] BackendRoomDisinviteRequest
- [x] BackendRoomUpdateRequest
- [x] BackendRoomDeleteRequest
- [x] BackendRoomInCallRequest
- [x] BackendRoomParticipantsRequest

Resolves #1173

### DIFF
--- a/api/signaling.go
+++ b/api/signaling.go
@@ -575,6 +575,7 @@ const (
 	ServerFeatureServerInfo            = "serverinfo"
 	ServerFeatureChatRelay             = "chat-relay"
 	ServerFeatureTransientSessionData  = "transient-sessiondata"
+	ServerFeatureChangedUsers          = "changed-users"
 
 	// Features to send to internal clients only.
 	ServerFeatureInternalVirtualSessions = "virtual-sessions"
@@ -601,6 +602,7 @@ var (
 		ServerFeatureServerInfo,
 		ServerFeatureChatRelay,
 		ServerFeatureTransientSessionData,
+		ServerFeatureChangedUsers,
 	}
 	DefaultFeaturesInternal = []string{
 		ServerFeatureInternalVirtualSessions,
@@ -617,6 +619,7 @@ var (
 		ServerFeatureServerInfo,
 		ServerFeatureChatRelay,
 		ServerFeatureTransientSessionData,
+		ServerFeatureChangedUsers,
 	}
 	DefaultWelcomeFeatures = []string{
 		ServerFeatureAudioVideoPermissions,
@@ -634,6 +637,7 @@ var (
 		ServerFeatureServerInfo,
 		ServerFeatureChatRelay,
 		ServerFeatureTransientSessionData,
+		ServerFeatureChangedUsers,
 	}
 )
 
--- a/api/signaling.go
+++ b/api/signaling.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"maps"
 	"net"
 	"net/url"
 	"slices"
@@ -1179,13 +1180,62 @@ type InternalServerMessage struct {
 
 // Type "event"
 
+type UserData = StringMap
+
+func (d UserData) Clone() UserData {
+	return maps.Clone(d)
+}
+
+func (d UserData) SessionId() (PublicSessionId, bool) {
+	return GetStringMapString[PublicSessionId](StringMap(d), "sessionId")
+}
+
+type UserDataMap map[PublicSessionId]UserData
+
+func (m UserDataMap) Users() UserDataList {
+	if len(m) == 0 {
+		return nil
+	}
+
+	return slices.Collect(maps.Values(m))
+}
+
+type UserDataList []UserData
+
+func (l UserDataList) Clone() UserDataList {
+	if len(l) == 0 {
+		return nil
+	}
+
+	clone := make(UserDataList, len(l))
+	for idx, u := range l {
+		clone[idx] = u.Clone()
+	}
+	return clone
+}
+
+func (l UserDataList) Map() UserDataMap {
+	if len(l) == 0 {
+		return nil
+	}
+
+	result := make(UserDataMap, len(l))
+	for _, d := range l {
+		sid, found := d.SessionId()
+		if found {
+			result[sid] = d
+		}
+	}
+	return result
+}
+
 type RoomEventServerMessage struct {
 	RoomId     string          `json:"roomid"`
 	Properties json.RawMessage `json:"properties,omitempty"`
 	// TODO(jojo): Change "InCall" to "int" when #914 has landed in NC Talk.
 	InCall  json.RawMessage `json:"incall,omitempty"`
-	Changed []StringMap     `json:"changed,omitempty"`
-	Users   []StringMap     `json:"users,omitempty"`
+	Changed UserDataList    `json:"changed,omitempty"`
+	Users   UserDataList    `json:"users,omitempty"`
 
 	All bool `json:"all,omitempty"`
 }
--- a/api/signaling.go
+++ b/api/signaling.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"maps"
 	"net"
 	"net/url"
 	"slices"
@@ -1181,13 +1182,62 @@ type InternalServerMessage struct {
 
 // Type "event"
 
+type UserData = StringMap
+
+func (d UserData) Clone() UserData {
+	return maps.Clone(d)
+}
+
+func (d UserData) SessionId() (PublicSessionId, bool) {
+	return GetStringMapString[PublicSessionId](StringMap(d), "sessionId")
+}
+
+type UserDataMap map[PublicSessionId]UserData
+
+func (m UserDataMap) Users() UserDataList {
+	if len(m) == 0 {
+		return nil
+	}
+
+	return slices.Collect(maps.Values(m))
+}
+
+type UserDataList []UserData
+
+func (l UserDataList) Clone() UserDataList {
+	if len(l) == 0 {
+		return nil
+	}
+
+	clone := make(UserDataList, len(l))
+	for idx, u := range l {
+		clone[idx] = u.Clone()
+	}
+	return clone
+}
+
+func (l UserDataList) Map() UserDataMap {
+	if len(l) == 0 {
+		return nil
+	}
+
+	result := make(UserDataMap, len(l))
+	for _, d := range l {
+		sid, found := d.SessionId()
+		if found {
+			result[sid] = d
+		}
+	}
+	return result
+}
+
 type RoomEventServerMessage struct {
 	RoomId     string          `json:"roomid"`
 	Properties json.RawMessage `json:"properties,omitempty"`
 	// TODO(jojo): Change "InCall" to "int" when #914 has landed in NC Talk.
 	InCall  json.RawMessage `json:"incall,omitempty"`
-	Changed []StringMap     `json:"changed,omitempty"`
-	Users   []StringMap     `json:"users,omitempty"`
+	Changed UserDataList    `json:"changed,omitempty"`
+	Users   UserDataList    `json:"users,omitempty"`
 
 	All bool `json:"all,omitempty"`
 }
--- a/api/signaling.go
+++ b/api/signaling.go
@@ -577,6 +577,7 @@ const (
 	ServerFeatureServerInfo            = "serverinfo"
 	ServerFeatureChatRelay             = "chat-relay"
 	ServerFeatureTransientSessionData  = "transient-sessiondata"
+	ServerFeatureChangedUsers          = "changed-users"
 
 	// Features to send to internal clients only.
 	ServerFeatureInternalVirtualSessions = "virtual-sessions"
@@ -603,6 +604,7 @@ var (
 		ServerFeatureServerInfo,
 		ServerFeatureChatRelay,
 		ServerFeatureTransientSessionData,
+		ServerFeatureChangedUsers,
 	}
 	DefaultFeaturesInternal = []string{
 		ServerFeatureInternalVirtualSessions,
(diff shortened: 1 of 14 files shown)


nickvergessen:
@fancycode to not break, the Talk code would need to know whether the HPB has this feature. Can you add a new feature header?

fancycode:
> @fancycode to not break, the Talk code would need to know whether the HPB has this feature. Can you add a new feature header?

Rebased and added as `changed-users`.

If we add more features in the future that are only relevant for the Talk backend, I should split them up so the clients don't receive feature ids they don't need.

danxuliu (commented):
A couple of issues:
- If in a conversation there is only one active participant and that participant starts a call, the client receives a `participants->update` signaling message with the participant duplicated in `users`, although in one case with `inCall: 0` and in the other `inCall: 7`
- The signaling server now keeps track of the participants to include them in `participants->update` when they are in a call. However, `participants->update` will include any active participant whose `inCall` changed at some point, even after they left the call (when `inCall` becomes 0 it should be included, but it should not be necessary to keep including them in further messages if `inCall` is still 0).
  - The participants will be included also if the `participants->update` is unrelated to a call. For example, if the permissions of a participant change the signaling message will include not only the participant whose permissions changed, but any other participant that are or were at some point in a call.
  - These extra/unneeded participants should™ not be problematic, but it caught my eye and I wondered if it was intended/expected.

nickvergessen:
Talk side is merged, and toggles behaviour based on the capability.
I guess the new HPB also still handles the old behaviour, as we did only backport to stable34.

But then this would be good to get merged and release (or test packaged) so we can put it live? :)

fancycode:
> A couple of issues:
> 
> * If in a conversation there is only one active participant and that participant starts a call, the client receives a `participants->update` signaling message with the participant duplicated in `users`, although in one case with `inCall: 0` and in the other `inCall: 7`

I saw this once yesterday while testing but can't reproduce any more. A user should only be included once and only with the latest properties. Does this happen every time for you?

> * The signaling server now keeps track of the participants to include them in `participants->update` when they are in a call. However, `participants->update` will include any active participant whose `inCall` changed at some point, even after they left the call (when `inCall` becomes 0 it should be included, but it should not be necessary to keep including them in further messages if `inCall` is still 0).
>   * The participants will be included also if the `participants->update` is unrelated to a call. For example, if the permissions of a participant change the signaling message will include not only the participant whose permissions changed, but any other participant that are or were at some point in a call.
>   * These extra/unneeded participants should™ not be problematic, but it caught my eye and I wondered if it was intended/expected.

This is from my [comment above](https://github.com/strukturag/nextcloud-spreed-signaling/pull/1175#issuecomment-3754713624), for backwards compatibility:

>> Also the signaling server still includes all participants in the updates events it sends out to connected sessions triggered by BackendRoomInCallRequest / BackendRoomParticipantsRequest. We should get rid of this too, if all clients (web/mobile/desktop) support receiving only changed participants.

danxuliu:
> > A couple of issues:
> > 
> > * If in a conversation there is only one active participant and that participant starts a call, the client receives a `participants->update` signaling message with the participant duplicated in `users`, although in one case with `inCall: 0` and in the other `inCall: 7`
> 
> I saw this once yesterday while testing but can't reproduce any more. A user should only be included once and only with the latest properties. Does this happen every time for you?

Ah, it happens when starting a call, leaving the call, reloading the conversation and starting a call again. So the duplicated user with `inCall: 0` is actually the previous session.

> > * The signaling server now keeps track of the participants to include them in `participants->update` when they are in a call. However, `participants->update` will include any active participant whose `inCall` changed at some point, even after they left the call (when `inCall` becomes 0 it should be included, but it should not be necessary to keep including them in further messages if `inCall` is still 0).
> >   
> >   * The participants will be included also if the `participants->update` is unrelated to a call. For example, if the permissions of a participant change the signaling message will include not only the participant whose permissions changed, but any other participant that are or were at some point in a call.
> >   * These extra/unneeded participants should™ not be problematic, but it caught my eye and I wondered if it was intended/expected.
> 
> This is from my [comment above](https://github.com/strukturag/nextcloud-spreed-signaling/pull/1175#issuecomment-3754713624), for backwards compatibility:
> 
> > > Also the signaling server still includes all participants in the updates events it sends out to connected sessions triggered by BackendRoomInCallRequest / BackendRoomParticipantsRequest. We should get rid of this too, if all clients (web/mobile/desktop) support receiving only changed participants.

What I had understood (and expected) from that comment was that the signaling server would keep track of the participants in a conversation and send them in `participants -> update`, even if they are no longer sent by the Nextcloud server. So if you modify the permissions of a participant then the `participants -> update` signaling message still includes all the (active, see `Previously existing issues` in https://github.com/nextcloud/spreed/pull/16759) participants in the conversation.

However, the signaling server only seems to keep track of the participants that were active in a call. So in the mentioned example `participants -> update` includes the participant whose permissions changed and all the participants that, at some point, were active in a call (and are still active in the conversation). If the permissions of another participant are modified then `participants -> update` includes this new participant, but not the previous one, and all the participants that at some point were active in a call. That behaviour is a bit strange, so I was not sure if it was the expected one :-)

Similarly, when the Nextcloud server provided the changed participants in a call, the `participants -> update` signaling messages sent by the signaling server included the participants currently active in the call, including the participants that just left. However, it did not include participants that, at some point, were in the call, as it happens now.

As mentioned before those extra participants should not be problematic but, at the same time, they should not be needed either. Having said that, if that behaviour is intended/expected, then no problem :+1:

fancycode:
> > > A couple of issues:
> > > 
> > > * If in a conversation there is only one active participant and that participant starts a call, the client receives a `participants->update` signaling message with the participant duplicated in `users`, although in one case with `inCall: 0` and in the other `inCall: 7`
> > 
> > 
> > I saw this once yesterday while testing but can't reproduce any more. A user should only be included once and only with the latest properties. Does this happen every time for you?
> 
> Ah, it happens when starting a call, leaving the call, reloading the conversation and starting a call again. So the duplicated user with `inCall: 0` is actually the previous session.

Still can't reproduce using these steps. When the conversation is reloaded (assuming a reload in the browser), the session is removed and the previous user entry gets removed:
https://github.com/strukturag/nextcloud-spreed-signaling/pull/1175/files#diff-917a377d5c8189edf91faf946df931fb415e8d74be6897e4d52c478efa73e2daR577

Even if I manually delay the backend request from Talk to set `inCall: 0` for the previous session, so that it arrives after the new session has already reconnected, the signaling server still doesn't send a `participants -> update` for the previous session.

> > > * The signaling server now keeps track of the participants to include them in `participants->update` when they are in a call. However, `participants->update` will include any active participant whose `inCall` changed at some point, even after they left the call (when `inCall` becomes 0 it should be included, but it should not be necessary to keep including them in further messages if `inCall` is still 0).
> > >   
> > >   * The participants will be included also if the `participants->update` is unrelated to a call. For example, if the permissions of a participant change the signaling message will include not only the participant whose permissions changed, but any other participant that are or were at some point in a call.
> > >   * These extra/unneeded participants should™ not be problematic, but it caught my eye and I wondered if it was intended/expected.
> > 
> > 
> > This is from my [comment above](https://github.com/strukturag/nextcloud-spreed-signaling/pull/1175#issuecomment-3754713624), for backwards compatibility:
> > > > Also the signaling server still includes all participants in the updates events it sends out to connected sessions triggered by BackendRoomInCallRequest / BackendRoomParticipantsRequest. We should get rid of this too, if all clients (web/mobile/desktop) support receiving only changed participants.
> 
> What I had understood (and expected) from that comment was that the signaling server would keep track of the participants in a conversation and send them in `participants -> update`, even if they are no longer sent by the Nextcloud server. So if you modify the permissions of a participant then the `participants -> update` signaling message still includes all the (active, see `Previously existing issues` in [nextcloud/spreed#16759](https://github.com/nextcloud/spreed/pull/16759)) participants in the conversation.
> 
> However, the signaling server only seems to keep track of the participants that were active in a call. So in the mentioned example `participants -> update` includes the participant whose permissions changed and all the participants that, at some point, were active in a call (and are still active in the conversation). If the permissions of another participant are modified then `participants -> update` includes this new participant, but not the previous one, and all the participants that at some point were active in a call. That behaviour is a bit strange, so I was not sure if it was the expected one :-)

So you mean, the `participants -> update` event only contains information about connected users that had their permissions changed but did not join the call if triggered by a backend request from Talk, but not in `participants -> update` events generated at a later time? I will look into this.

> Similarly, when the Nextcloud server provided the changed participants in a call, the `participants -> update` signaling messages sent by the signaling server included the participants currently active in the call, including the participants that just left. However, it did not include participants that, at some point, were in the call, as it happens now.
> 
> As mentioned before those extra participants should not be problematic but, at the same time, they should not be needed either. Having said that, if that behaviour is intended/expected, then no problem 👍

As commented above, user information for sessions that were deleted should not be included in future `participants -> update` events because their entry is removed once the session is removed from the room.

danxuliu:
> > Ah, it happens when starting a call, leaving the call, reloading the conversation and starting a call again. So the duplicated user with `inCall: 0` is actually the previous session.
> 
> Still can't reproduce using these steps. When the conversation is reloaded (assuming a reload in the browser), the session is removed and the previous user entry gets removed: https://github.com/strukturag/nextcloud-spreed-signaling/pull/1175/files#diff-917a377d5c8189edf91faf946df931fb415e8d74be6897e4d52c478efa73e2daR577
> 
> Even if I manually delay the backend request from Talk to set `inCall: 0` for the previous session, so that it arrives after the new session has already reconnected, the signaling server still doesn't send a `participants -> update` for the previous session.

I think that I found what is going on. It depends on the browser:

- When reloading in Chromium:
```
hub.go:1162: Unregister XXX (private=YYY)
clientsession.go:601: Removed room session ZZZ: map[room:map[permissions:[publish-audio publish-video publi...
```

- When reloading in Firefox:
```
hub.go:1162: Unregister XXX (private=YYY)
```
and later
```
hub.go:860: Closing expired session XXX (private=YYY)
clientsession.go:601: Removed room ZZZ: map[error:map[code:no_such_room message:The user is not invited to this room.] type:error]
```

It seems that Chromium sends the `Bye` message before closing the WebSocket connection, while Firefox closes it "too fast". Therefore, after reloading in Firefox the previous session was not removed yet, and thus still appeared in the tracked participants.

In any case, this was indirectly fixed (I mean, fixing this specific case was not the main goal) in the commits that I added, see below.
 
> So you mean, the `participants -> update` event only contains information about connected users that had their permissions changed but did not join the call if triggered by a backend request from Talk, but not in `participants -> update` events generated at a later time? I will look into this.

Correct, but... after thinking more about it I came to the conclusion that this is not needed :-) Before the changes in this pull request the `participants->update` message included either all the active participants in the conversation or all the participants in the call (which is a subset of the former). With the new code it is possible to still provide all the participants in the call, but it is not possible to provide all the active participants in the conversation as before.

Due to the previous behaviour the clients assumed that, at least, all the participants in the call will be always included in those messages, but they have no expectations regarding other participants in the conversation. Therefore, I think that it is not necessary to keep track of modified participants that are not in the call, as that will not provide the same behaviour as before, and given that there will be a behaviour change anyway there is no need to keep track of them either given that the clients will ignore them (they are still relevant and used when they are modified, but not afterwards).

Due to that I pushed some additional commits in https://github.com/danxuliu/nextcloud-spreed-signaling/commits/remove-all-users-talk-requests/ (I do not have permissions to push directly to this repository).

For further details please refer to the commit message of [`Do not keep track of modified participants if they did not join a call`](https://github.com/danxuliu/nextcloud-spreed-signaling/commit/10941733f8411b42df212377136125bfad57d44c).

> As commented above, user information for sessions that were deleted should not be included in future `participants -> update` events because their entry is removed once the session is removed from the room.

Yes, but they were still included in every `participants->update` message sent after they left the call and before they left the room. As mentioned above I think that is not necessary, so I added another commit that still includes them when the leave the call but not anymore afterwards (the same behaviour as before). For further details please refer to the commit message of [`Do not keep track of participants after they left the call`](https://github.com/danxuliu/nextcloud-spreed-signaling/commit/f2b12087c4a3b87337910116da26fde10bc49c05)

Finally, the last commit [`Remove tracked participants when the call is left for all`](https://github.com/danxuliu/nextcloud-spreed-signaling/commit/18a6ce986d7278db84138bfe02670da646a4f335) just fixes the participant list provided by the signaling server in some cases after the call was ended for everyone. This was actually a bug in the previous code, even if it was a corner case.

PR #1202, opened 2026-02-15
docker: pin spreedbackend uid and add user group

lnobach:
I would suggest to pin the uid of the docker process to a specific number. The added group is a nice-to-have, e.g. if you want the spreedbackend process to access files by giving the files a specific gid.


fancycode:
Good idea, could you change this also for the signaling proxy container so they are in sync? Thanks.

fancycode:
Thanks!

PR #1208, opened 2026-02-26
Don't limit size of received Janus events.

fancycode:
@danxuliu FYI


danxuliu (commented):
Sorry for the delay, some long overdue notes :facepalm:

Media stats from Janus seems to be ~1000 bytes per media report/event. There are separate reports for audio and video connections for each participant, but the size of each report is roughly the same for publishers and subscribers.

By default each report is sent every second, it can be configured with [`events->stats_period` in _janus.jcfg_](https://github.com/meetecho/janus-gateway/blob/0c29e0e31665092ac540eacb9c1b0e912e36ea84/conf/janus.jcfg.sample.in#L480). Obviously with a larger period there less data is sent.

Although each report is sent at rate X several reports are "batched" and sent together; the frequency of these batches does not seem to be configurable, they are just sent when possible and the available reports at that time are included. Therefore the batches may contain a single report or tens of them.

In any case, by default [at most 100 of them are grouped together](https://github.com/meetecho/janus-gateway/blob/766420f724bc7ae2a2d8bf92c5d9e4cf071567b1/src/events/janus_wsevh.c#L650) (it is possible to configure whether to group them or not, but not the maximum number of reports to group), so the WebSocket message should be, at most, around ~100KB (and in actual tests I never saw it above that size). If there are more reports to be sent than 100 then they are sent in a new message. So even if now the size of WebSocket messages received from Janus is unlimited in practice they should still have a reasonable size :-)

PR #1248, opened 2026-05-11
feat: added github pipeline for janus

creyD:
Thanks for the hard work! I am using this signaling server for a while now and created this build pipeline, when I started: https://github.com/creyD/janus for just creating a ghcr.io container image to pull in my deployments (as I didn't find it anywhere else prebuilt).

Maybe this is interesting for you, or maybe you have your reasons for not publishing this on your own in the first place. If there is anything to change or test, feel free to ping me.

### DIFF
--- a/.github/workflows/docker-janus.yml
+++ b/.github/workflows/docker-janus.yml
@@ -14,9 +14,11 @@ on:
 
 permissions:
   contents: read
+  packages: write
 
 env:
   TEST_TAG: strukturag/nextcloud-spreed-signaling:janus-test
+  IMAGE_TAG: ghcr.io/strukturag/nextcloud-spreed-signaling:janus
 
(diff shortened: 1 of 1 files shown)


creyD:
@fancycode Are you the maintainer of this project? Are you interested in merging this?

fancycode:
Yes, I'm a maintainer of the signaling server. 

That said, I don't think it's a good idea to build and publish Janus container images from this repository here. Janus is an external project, so versioning / release timelines are totally different from the signaling server.

It should be in a separate project, so container image releases can follow Janus releases.

The Dockerfile in `docker/janus` was provided by the community and is only kept in the repository here as an example for users on how to build / run Janus themselves. It's part of the CI to detect problems with potential future PRs.

PR #1256, opened 2026-06-02
Fix internal incall flags of users if updated with "all: true"

fancycode:
@SystemKeeper FYI


SystemKeeper:
Tested and fixed the issue 🥳

PR #1263, opened 2026-06-03
Fix race between publisher connected / subscriber waiting.

### LINKED ISSUE #1257: Race condition: subscriber misses publisher-connected notification → "No such feed" on reconnect
## Summary

During simultaneous reconnections (e.g. after a proxy timeout or network interruption), subscribers intermittently fail to receive audio/video from publishers. Janus reports \`No such feed (N)\` and the subscriber remains silently connected but unable to hear the publisher, despite the call appearing active on both sides.

## Environment

- nextcloud-spreed-signaling 2.1.1
- Janus WebRTC Gateway 1.4.0
- Nextcloud Talk 23.0.5 / Nextcloud 33.0.4
- ~30 concurrent users, multiple Talk Desktop clients (v2.x)
- Production deployment, single-server setup

## Steps to reproduce

1. Multiple clients connect to the same room
2. A reconnect event occurs simultaneously for several clients (e.g. proxy timeout, brief network interruption, or HPB restart)
3. Publisher reconnects and sends `webrtcup` notification
4. Subscriber reconnects slightly later and creates its waiter *after* the notification has already fired

**Result:** The subscriber's `Wait()` call never returns. Subscriber eventually times out with "No such feed" from Janus VideoRoom. The participant cannot hear the publisher until they manually rejoin.

## Root cause

The bug is a classic lost-wakeup race in `async/notifier.go`:

```go
// notifier.go — simplified
func (n *Notifier) Notify(key string) {
    for _, w := range n.waiterMap[key] {
        close(w.ch)  // fires now
    }
}

func (n *Notifier) NewWaiter(key string) *Waiter {
    // If Notify() already fired before this call,
    // the channel is never closed → Wait() blocks forever
    w := &Waiter{ch: make(chan struct{})}
    n.waiterMap[key] = append(n.waiterMap[key], w)
    return w
}
```

In `sfu/janus/subscriber.go`, the waiter is created **once before** the retry loop:

```go
waiter := newPublisherConnectedWaiter(...)
for retry := 0; retry < maxRetries; retry++ {
    select {
    case <-waiter.ch:  // already closed on first iteration → busy-loop on retry
    }
}
```

This causes two related problems:
1. If the publisher notification fires before `NewWaiter()` is called → the notification is permanently lost → subscriber blocks indefinitely
2. After the sticky-notifier fix, re-using the same (already-closed) waiter in the retry loop causes a busy-loop

## Production impact

In our deployment, this race triggered during reconnect storms caused by proxy timeouts (all clients connected at similar times would disconnect simultaneously):

| Metric | Before fix | After fix (5 days) |
|--------|-----------|-------------------|
| "No such feed" errors | 10–44 per event | 0 |
| Events per day | ~10 | 0 |
| Users affected per event | 2–5 | 0 |

A typical incident: one participant reconnects slightly ahead of others → publisher sends `webrtcup` → remaining subscribers create their waiters *after* the notification → entire room loses audio until timeout.

## Fix

PR #1254 proposes a two-part fix:

1. **`async/notifier.go`** — "sticky" notifications with 30-second TTL: `NewWaiter()` checks if a notification fired recently and returns a pre-closed channel immediately instead of blocking.

2. **`sfu/janus/subscriber.go`** — waiter is created inside the retry loop, ensuring each retry gets a fresh waiter and `stop()` is always called.

Both changes use only standard Go concurrency primitives (channel closing as broadcast, `time.Since()` for TTL check). No external or undocumented APIs.

## Related

- PR #1254 (fix proposal with patch)
- Issue #958 (related but different scenario: publisher room destroyed while waiting)

fancycode:
Alternative implementation to #1254, fixes #1257.

### DIFF
--- a/async/notifier.go
+++ b/async/notifier.go
@@ -23,29 +23,99 @@ package async
 
 import (
 	"context"
+	"errors"
 	"sync"
+	"sync/atomic"
 )
 
-type rootWaiter struct {
+var (
+	ErrDuplicateSignaler = errors.New("duplicate signaler") // +checklocksignore: Global readonly variable.
+	ErrAlreadyReleased   = errors.New("already released")
+)
+
+type notifierEntry struct {
+	sync.Mutex
+
+	ref atomic.Int32
 	key string
 	ch  chan struct{}
+
+	// +checklocks:Mutex
+	signaler *Signaler
+	// +checklocks:Mutex
+	waiter map[*Waiter]bool
 }
 
-func (w *rootWaiter) notify() {
-	close(w.ch)
+func (w *notifierEntry) setSignaler(signaler *Signaler) {
+	w.Lock()
+	defer w.Unlock()
+
+	if w.signaler != nil {
+		panic(ErrDuplicateSignaler)
+	}
+
+	w.signaler = signaler
+}
+
+func (w *notifierEntry) clearSignaler(signaler *Signaler) {
+	w.Lock()
+	defer w.Unlock()
+
+	if w.signaler != signaler {
+		panic("unknown signaler")
+	}
+
+	w.signaler = nil
+}
+
+func (w *notifierEntry) addWaiter(waiter *Waiter) {
+	w.Lock()
+	defer w.Unlock()
+
+	w.waiter[waiter] = true
+}
+
+func (w *notifierEntry) removeWaiter(waiter *Waiter) {
+	w.Lock()
+	defer w.Unlock()
+
+	delete(w.waiter, waiter)
+}
+
+func (w *notifierEntry) notify() {
+	select {
+	case <-w.ch:
+		// Already closed
+	default:
+		close(w.ch)
+	}
 }
 
 type Waiter struct {
-	key string
-	ch  <-chan struct{}
+	entry atomic.Pointer[notifierEntry]
 }
 
 func (w *Waiter) Wait(ctx context.Context) error {
+	entry := w.entry.Load()
+	if entry == nil {
+		return nil
+	}
+
 	select {
-	case <-w.ch:
+	case <-entry.ch:
 		return nil
 	case <-ctx.Done():
 		return ctx.Err()
+	}
+}
+
+type Signaler struct {
+	entry atomic.Pointer[notifierEntry]
+}
+
+func (s *Signaler) Signal() {
+	if entry := s.entry.Load(); entry != nil {
+		entry.notify()
 	}
 }
 
@@ -53,43 +123,53 @@ type Notifier struct {
 	sync.Mutex
 
 	// +checklocks:Mutex
-	waiters map[string]*rootWaiter
-	// +checklocks:Mutex
-	waiterMap map[string]map[*Waiter]bool
+	waiters map[string]*notifierEntry
 }
 
 type ReleaseFunc func()
 
-func (n *Notifier) NewWaiter(key string) (*Waiter, ReleaseFunc) {
+func (n *Notifier) createEntry(key string) *notifierEntry {
 	n.Lock()
 	defer n.Unlock()
 
 	waiter, found := n.waiters[key]
 	if !found {
-		waiter = &rootWaiter{
-			key: key,
-			ch:  make(chan struct{}),
+		waiter = &notifierEntry{
+			key:    key,
+			ch:     make(chan struct{}),
+			waiter: make(map[*Waiter]bool),
 		}
 
 		if n.waiters == nil {
-			n.waiters = make(map[string]*rootWaiter)
-		}
-		if n.waiterMap == nil {
-			n.waiterMap = make(map[string]map[*Waiter]bool)
+			n.waiters = make(map[string]*notifierEntry)
 		}
 		n.waiters[key] = waiter
-		if _, found := n.waiterMap[key]; !found {
-			n.waiterMap[key] = make(map[*Waiter]bool)
-		}
 	}
 
-	w := &Waiter{
-		key: key,
-		ch:  waiter.ch,
-	}
-	n.waiterMap[key][w] = true
+	waiter.ref.Add(1)
+	return waiter
+}
+
+func (n *Notifier) NewSignaler(key string) (*Signaler, ReleaseFunc) {
+	entry := n.createEntry(key)
+
+	s := &Signaler{}
+	s.entry.Store(entry)
+	entry.setSignaler(s)
 	releaseFunc := func() {
-		n.release(w)
+		n.releaseSignaler(s)
+	}
+	return s, releaseFunc
+}
+
+func (n *Notifier) NewWaiter(key string) (*Waiter, ReleaseFunc) {
+	entry := n.createEntry(key)
+
+	w := &Waiter{}
+	w.entry.Store(entry)
+	entry.addWaiter(w)
+	releaseFunc := func() {
+		n.releaseWaiter(w)
 	}
 	return w, releaseFunc
 }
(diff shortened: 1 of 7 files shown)


vladopol:
Tested in production on our deployment (Nextcloud Talk 23.0.5, HPB 2.1.1, Janus 1.4.0, ~30 concurrent users, Talk Desktop 2.1.x on Windows/macOS).

Built from the `race-wait-for-publisher` branch and deployed replacing our previous patch (#1254). HPB started cleanly, all clients reconnected within seconds, no errors in logs.

Monitoring for regressions — will report back after a few days of normal usage.

vladopol:
Completed testing over ~1 hour with 3 test calls (2 participants each, one on internal network + Talk Desktop, one on iOS v23.0.2 via mobile/external network).

**Results:**

The `wait and retry` mechanism fired in every call and worked correctly:
```
subscriber.go:248: Publisher ... not sending yet for video, wait and retry
subscriber.go:261: Retry subscribing video
subscriber.go:104: Subscriber ... received connected  ✓
```

Both subscribers successfully received `connected` in each call. No race-condition "No such feed" errors with random feed IDs observed (only feed=1 noise from outdated Talk v1.1.7 clients — unrelated to this fix).

**Comparison with #1254:** Same observable outcome — zero race-condition errors under both patches. The reference-counting approach in #1263 is architecturally sounder (no TTL expiry risk) and more transparent in logs.

No regressions found. Will continue monitoring production over the next few days and report back.

vladopol:
**Production incident with 1263-test: "No such feed (1)" cascade not prevented**

We have been running `1263-test` in production since early June (Nextcloud Talk 23.0.5, spreed-signaling 2.1.1, Janus 1.4.0). Initial testing looked good — three back-to-back calls, zero `No such feed` errors, both participants received `Subscriber received connected`. We posted that result earlier in this thread.

However, we later observed a production incident that PR#1263 did not prevent.

**What happened**

At T+0: three VideoRoom sessions closed simultaneously, including one with a 43 MB recording. This created a burst of cleanup events on the single WebSocket connection between HPB and Janus.

At T+37s: a publisher (internal client, Talk Desktop v2.0.5) joined the room. Janus immediately logged:
```
[WARN] Didn't receive video (substream #1) for more than 1 second(s)
[WARN] Didn't receive video for more than 1 second(s)
```

At T+67s: a subscriber joined (external client using TURN/TLS relay).

At T+68s: Janus logged:
```
[ERR] [plugins/janus_videoroom.c:janus_videoroom_handler:10208] No such feed (1)
```

Note the feed ID: **1** — a default/uninitialized value, not the publisher's actual Janus feed ID. This indicates HPB initiated the subscriber join before it had received and processed the publisher's `joinroom` response from Janus.

**The cascade**

49 total `No such feed` errors over the following 10 minutes. Each error destroyed the subscriber handle and triggered a retry, which created a new ICE agent with new candidates, which then failed:

```
ICE failures/min: 16 → 47 → 32 → 48 → 49 → 78 → 90 → 31 → 64 → 8
```

At peak (90/min) Janus's message queue backed up enough that HPB's `requestoffer` calls hit context timeout:
```
Could not send MCU message {type:requestoffer}: context deadline exceeded
Could not send MCU message {type:requestoffer}: not connected
```

The cascade self-resolved after ~17 minutes. The subscriber eventually connected successfully.

**Root cause analysis**

There are two overlapping issues, both rooted in the same architectural constraint.

**The underlying constraint: single WebSocket, single Janus message processing thread**

All HPB↔Janus communication goes through one persistent WebSocket connection. Janus processes messages from that connection in a single thread. This means any burst of events — even from routine activity like three simultaneous session closures — creates a processing queue. Messages behind that queue are delayed regardless of server hardware capacity. Three simultaneous closures is already enough to open a race window; five or ten would make it proportionally wider and more likely to affect multiple subscribers simultaneously.

**Issue 1 — feed=1 race (root cause)**

HPB initiates the subscriber `join` before the publisher's `joinroom` response has arrived from Janus, sending `feed=1` (uninitialized) instead of the actual feed ID. Under normal load this window is negligible. Under any Janus processing pressure — even three simultaneous session closures — the `joinroom` response is delayed in the queue long enough for the race to trigger.

This is distinct from the `webrtcup` race addressed by this PR. The publisher is already registered in the VideoRoom but HPB does not yet know its feed ID.

**Issue 2 — tight retry loop (amplifier)**

Once `No such feed` fires, the retry loop runs without any delay between iterations. If the notification/waiter resolves immediately (publisher already connected), the loop becomes tight — retrying as fast as Janus can respond with another error. Each iteration creates a new ICE agent, generating candidates that then fail, flooding Janus further and worsening the queue backup that caused the problem in the first place.

**Suggestions**

**Fix 1 — wait for publisher feed ID before initiating subscriber join**

Before sending the subscriber `join` message, HPB should verify that the publisher's actual Janus feed ID has been received (i.e. the publisher's `joinroom` response from Janus has been processed). If not yet available, HPB should block with a waiter — analogous to the existing `webrtcup` waiter — rather than proceeding with an uninitialized value of `1`. This addresses the root cause directly.

**Fix 2 — exponential backoff in the retry loop**

In addition, adding backoff between `No such feed` retries in `subscriber.go` would limit the ICE agent storm when the error persists across multiple iterations:

```go
case janus.JANUS_VIDEOROOM_ERROR_NO_SUCH_FEED:
    retryCount++
    backoff := time.Duration(retryCount*retryCount) * 100 * time.Millisecond
    if backoff > 2*time.Second {
        backoff = 2*time.Second
    }
    waiter, stop := p.mcu.newPublisherConnectedWaiter(p.publisher, p.streamType)
    if err := waiter.Wait(ctx); err != nil {
        stop()
        ...
    }
    stop()
    select {
    case <-ctx.Done():
        callback(ctx.Err(), nil)
        return
    case <-time.After(backoff):
    }
    goto retry
```

This does not affect the happy path (first retry succeeds immediately when publisher is ready) but prevents the retry loop from amplifying a temporary Janus queue backup into a 17-minute ICE storm.

vladopol:
**3-week production update (June 22)**

We have been running the `1263-test` build (`c21c347`) in production for three weeks now (~30 concurrent users, Nextcloud Talk 23.0.5, Janus 1.4.0).

**Original race condition: resolved**

Zero occurrences of "No such feed" with random feed IDs (7, 15, 43, etc.) since June 3. The `wait and retry` mechanism continues to fire and recover correctly on every call. We consider the original race fully addressed by this PR.

**Revisiting the June 6 incident**

On reflection, the "No such feed (1)" cascade we reported on June 6 more likely originated from stale subscriber handles held by outdated Talk clients (v1.1.7) rather than HPB sending an uninitialized feed ID. We observe a persistent baseline of "No such feed" with feed IDs specifically equal to 1 or 2 across all our production logs — always correlated with the older client version, not with publisher join timing. We may have misattributed causation in that report.

That said, the **backoff suggestion** from our June 6 comment remains valid independently: a tight retry loop amplifies any "No such feed" storm regardless of its origin, and adding backoff would limit the blast radius.

**Summary**

PR #1263 is stable and effective on our deployment. Happy to retest against any updated revision if you push changes.

fancycode:
> That said, the backoff suggestion from our June 6 comment remains valid independently: a tight retry loop amplifies any "No such feed" storm regardless of its origin, and adding backoff would limit the blast radius.

The subscriber will wait for the "publisher connected" event before retrying:
https://github.com/strukturag/nextcloud-spreed-signaling/blob/c21c347ba935636c87dc0243b72529eb67d62249/sfu/janus/subscriber.go#L256

This event is only triggered once the publisher has created the room and the WebRTC connection is established (i.e. the `webrtcup` event from Janus):
https://github.com/strukturag/nextcloud-spreed-signaling/blob/c21c347ba935636c87dc0243b72529eb67d62249/sfu/janus/publisher.go#L100

At this point, the new subscribe request should succeed.

Not sure why you would need an additional exponential backoff.

fancycode:
Thanks for testing and your feedback! Merging for now, feel free to open a new issue if you have more findings.

Issue #1274, opened 2026-07-08
Remote (cross-server) subscriber deadlocks on re-attach: getPublisher never resolves a remote publisher


kalinbogatzevski:
# Remote (cross-server) subscriber deadlocks on re-attach: `getPublisher` never resolves a remote publisher

### Description

When a subscriber's publisher lives on a **different** signaling server — the
remote-streams / cross-proxy SFU path, with several Janus proxies listed in
`[mcu] url` — the subscriber deadlocks whenever it has to **re-attach** its
Janus handle. The re-attach blocks for the whole MCU request timeout and the
call is then torn down with `context deadline exceeded`.

### Steps to reproduce

1. Configure ≥2 signaling proxy servers with remote streams enabled
   (`hostname`, `token_id`, `token_key` in the proxy configuration) and list
   them in the signaling server's `[mcu] url`, so a publisher and a subscriber
   can be placed on different proxies.
2. Have two participants that land on **different** proxies join the same call
   (e.g. GeoIP-based placement across regions): A publishes on proxy-1, B
   subscribes from proxy-2 — B is a *remote* subscriber.
3. Trigger a subscriber re-attach, either:
   - the subscriber re-joins while its previous subscriber object is still
     connected → Janus returns `JANUS_VIDEOROOM_ERROR_ALREADY_JOINED`, or
   - `NotifyReconnected` fires (proxy reconnect).

### Actual behaviour

The re-attach blocks for the full MCU request timeout (`timeout` in the `[mcu]`
section, 10 seconds by default), then fails with `context deadline exceeded`;
the subscriber (and usually the call) is torn down. Logged e.g. as:
`already connected as subscriber for video, error during re-joining: context deadline exceeded`.

### Root cause

Both re-attach entry points route through
`janusSFU.getOrCreateSubscriberHandle` → `getPublisher`. `getPublisher`
resolves **local** publishers from `m.publishers` and blocks on a condition
variable until the context deadline if the publisher is not there. A remote
publisher only ever exists in `m.remotePublishers`, so for a remote subscriber
the wait can never be satisfied — it always runs into the context deadline.

The two entry points:
- `janusSubscriber.joinRoom`, the `JANUS_VIDEOROOM_ERROR_ALREADY_JOINED` branch
  (`sfu/janus/subscriber.go`).
- `janusRemoteSubscriber.NotifyReconnected` (`sfu/janus/remote_subscriber.go`).

[...]

PR #1275, opened 2026-07-08
Fix deadlock when a remote (cross-server) subscriber re-attaches its handle

kalinbogatzevski:
### Problem

A **remote subscriber** — one whose publisher lives on another signaling server
(the remote-streams SFU path) — deadlocks when it has to re-attach its Janus
handle, either on `JANUS_VIDEOROOM_ERROR_ALREADY_JOINED` during a re-join or
from `NotifyReconnected`. Both paths call `getOrCreateSubscriberHandle` →
`getPublisher`, which resolves only **local** publishers from `m.publishers` and
blocks on a condition variable until the request context deadline. A remote
publisher lives only in `m.remotePublishers`, so the lookup can never succeed:
the subscriber stalls for the full MCU timeout and is torn down with
`context deadline exceeded`, dropping the call. Every re-attach of a remote
subscriber hits this; subscribers are remote whenever publisher and subscriber
are placed on different proxies (e.g. GeoIP-based placement across regions).

Details in #1274.

### Fix

`janusSubscriber` gets an optional `attachHandle` hook. When set, the
`ALREADY_JOINED` re-join path uses it instead of `getOrCreateSubscriberHandle`.
Remote subscribers install `attachRemoteHandle`, which attaches a fresh handle
directly from the already-known `*janusRemotePublisher` via a new
`getOrCreateRemoteSubscriberHandle` — never touching `getPublisher`.
`NotifyReconnected` uses the same helper. Local subscribers are completely
unchanged (the hook is nil for them).

The hook follows the function-field dispatch `janusClient` already uses
(`handleEvent`, `handleHangup`, …) and is assigned in `NewRemoteSubscriber`.
This mirrors what `NewRemoteSubscriber` already does for the **initial** attach.

### Testing

- `go build ./...`, `go vet ./sfu/janus/...` and `go test ./sfu/janus/...` all
  pass on `master` + this change.
- Added `Test_JanusRemoteSubscriberAlreadyJoined` and
  `Test_JanusRemoteSubscriberReconnect`, which **fail on current master** (the
  subscriber blocks in `getPublisher` until the MCU timeout and is closed) and
  **pass with this change**. The mock gateway needed a small extension so a room
  fed via `add_remote_publisher` is joinable by subscribers (sets
  `room.publisher` and an offer SDP).
- Also verified end-to-end on a 3-proxy deployment with GeoIP placement across
  regions: before, a cross-region subscriber reliably logged
  `context deadline exceeded` on re-attach and the call dropped; after, the
  re-attach resolves and the `ALREADY_JOINED` re-join completes with zero
  deadline errors.

### Notes

Remote streams is a preview feature (since #708). This change only touches the
remote-subscriber re-attach paths; the local subscriber path is untouched. The
touched files are identical between v2.1.1 and `master`, so this also
backports cleanly to the 2.1.x branch.

### DIFF
--- a/sfu/janus/janus.go
+++ b/sfu/janus/janus.go
@@ -956,6 +956,25 @@ func (m *janusSFU) getOrCreateSubscriberHandle(ctx context.Context, publisher ap
 	return handle, pub, nil
 }
 
+// getOrCreateRemoteSubscriberHandle attaches a fresh subscriber handle for a
+// remote publisher. Unlike getOrCreateSubscriberHandle it must not go through
+// getPublisher: remote publishers are tracked only in m.remotePublishers, so a
+// lookup in m.publishers would block until the context deadline.
+func (m *janusSFU) getOrCreateRemoteSubscriberHandle(ctx context.Context, pub *janusRemotePublisher) (*janus.Handle, error) {
+	session := m.session
+	if session == nil {
+		return nil, sfu.ErrNotConnected
+	}
+
+	handle, err := session.Attach(ctx, pluginVideoRoom)
+	if err != nil {
+		return nil, err
+	}
+
+	m.logger.Printf("Attached remote subscriber to room %d of publisher %s on handle %d", pub.roomId, pub.id, handle.Id)
+	return handle, nil
+}
+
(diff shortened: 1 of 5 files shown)


fancycode:
Thanks!

PR #1303, opened 2026-08-20
Ignore stale responses to superseded "join room" requests

kalinbogatzevski:
A session can be left in a room it already left, if the backend responses to two
"join room" requests complete out of order.

### How two joins overlap for one session

Messages from a single client are processed one at a time (`Client.processMessages`
reads them from a channel and dispatches them synchronously), so this is not about
a client sending two "room" messages in a row. The overlap comes from the session
outliving the client that started the request:

* `ClientSession.SetClient` replaces the client of an existing session, so a
  resumed client can start a new join while a previous one is still running.
* `remoteGrpcClient` feeds messages for a session from its own goroutine.

`Hub.processRoom` performs the backend request with `session.Context()` and a
timeout of `backendTimeout`, so it keeps running after the client that sent it is
gone. The response can therefore arrive after a newer join for the same session
has already completed.

`processJoinRoom` then applies it regardless: it calls `session.LeaveRoom(true)`
and joins the stale room. The session ends up in the room it left, and the room
session created by the newer join is removed again.

### Change

Each "join room" request takes a sequence number from the session before the
backend request is started (`ClientSession.NextJoinRoomSeq`). `processJoinRoom`
compares it against the current value before touching the session and returns
early if a newer request has been started since. The newer request applies its
own result, so nothing is lost by dropping the superseded one.

The check sits after the existing error handling and before `LeaveRoom`, so a
stale response no longer has any side effect. Two atomic operations on the normal
path, no locking.

### Test

`TestJoinRoomStaleResponseAfterClientSwitch` starts a join whose backend response
is delayed, closes the client, resumes the session on a second client, joins a
different room there, and then checks which room the session is in once the
delayed response has been processed.

It uses a new `test-room-very-slow` case in the test backend (500ms) rather than
the existing `test-room-slow` (100ms), so the resume and the second join have room
to complete without the timing being tight.

Without the change the test fails, and the log shows the stale response being
applied 389ms after the correct one:

```
joined room test-room
joined room test-room-very-slow      <- stale response
Removed room session test-room-...
```

### Checks

`go build ./...`, `go vet ./server/`, `gofmt -l server/` are clean, and
`go test ./server/` passes with and without `-race`.

Note that `go test -race -run "TestJoinRoom" ./server/` reports data races in
`api.NewWelcomeServerMessage` (a `slices.Sort` on package-level state, reached
through `NewHub` from several parallel tests). That is present on master without
this change as well, so it is left alone here.

### DIFF
--- a/server/clientsession.go
+++ b/server/clientsession.go
@@ -91,6 +91,7 @@ type ClientSession struct {
 	client       ClientWithSession
 	room         atomic.Pointer[Room]
 	roomJoinTime atomic.Int64
+	joinRoomSeq  atomic.Uint64
 	federation   atomic.Pointer[FederationClient]
 
 	roomSessionIdLock sync.RWMutex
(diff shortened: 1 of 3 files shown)


kalinbogatzevski:
The two failing `test` jobs are `TestRoom_Delete`, not the test added here:

```
--- FAIL: TestRoom_Delete (10.11s)
    testclient_test.go:687: error while waiting for closed connection
        context deadline exceeded
```

The same failure with the same 10.11s timeout happens on master without this
change, e.g. run 32342342314 (`go (1.27)`, master, earlier today):

```
--- FAIL: TestRoom_Delete (10.11s)
```

Locally `go test -count=10 -run TestRoom_Delete ./server/` passes, and the full
`./server/` package passes with and without `-race` on this branch. I do not have
permission to restart the jobs here.

